PR #2, opened 2025-02-02
docs: update README.md

eltociear:
alredy -> already


faiface:
Good catch, thanks! Merging

PR #5, opened 2025-02-03
Interaction combinator compilation & runtime

FranchuFranchu:
This PR adds support for compilation to interaction combinators (#3). It features a small (and not performant at all) interaction combinator runtime, which I took from another project I wrote, and a playground interface to use it. Note that the playground interface is read-only; you can't write your own interaction combinator nets and have the playground parse them.

Right now, it's still very incomplete, because there are many things that we haven't decided yet. It doesn't support `Choose` and `Either` with signals that aren't named `left` or `right`. It doesn't support `Break` and `Loop`. It doesn't handle captures at all, because I'm not sure what they are useful for yet. It doesn't handle incorrect code gracefully. I haven't written tests either.

However, this shows that it's possible. It was very easy to code this using #4. 

We need to decide if we want this after an type checker, or before. I believe finishing the implementation after a type checker is better. IC compilation without a type system is quite limited. For example, we won't be able to readback interaction combinator nets if we don't know their type.

I'm making this a draft because the complete design and plan for this hasn't been worked on yet.


faiface:
Impressive! Thanks a lot for putting this together!

You are probably right that it makes the most sense to complete this after a type system is implemented.

I already have a type system design in mind (with some questions open and up for refinement), so I'll put together an issue soon describing the design.

It seems to me that the type system is the next logical step. Initially, I thought replicable values were the next step, but since an implementation of those may be influenced by interaction nets/combinators, it may make sense to do them afterwards. What do you think?

But regarding interaction nets, I still have this question: What is the advantage of compiling to the combinators directly as opposed to compiling to custom interactio net nodes and rules? @FranchuFranchu

FranchuFranchu:
@faiface 

> It seems to me that the type system is the next logical step. Initially, I thought replicable values were the next step, but since an implementation of those may be influenced by interaction nets/combinators, it may make sense to do them afterwards. What do you think?

Compilation of replicable values is not very surprising. Compilation of weakening, dereliction, and contraction is very simple. Even if it's left implicit (I personally think making them explicit is better, though), it's definitely still doable, as long as we know which names refer to replicable values. 

Compilation of !-introduction does require more work from part of the compiler. An expression variant that builds a new replicable value, which also takes in a list of captured replicable values is an example of an approach that would be easy to work with. Example:

```
chan user |other_replicable_value| {
  let make_pair_with = replicable |other_replicable_value| {
       [a] (other_replicable_value) a
  }
  // the syntax for these operations can be anything else since they're quite simple to compile.
  // Contraction
  dup mpw1 mpw2 mpw3 = make_pair_with
  // Weakening
  era mpw3
  // Dereliction
  dere mpw1 = mpw1
  dere mpw2 = mpw2
  user(mpw1)
  user(mpw2)
  user!
}
```
 
> But regarding interaction nets, I still have this question: What is the advantage of compiling to the combinators directly as opposed to compiling to custom interaction net nodes and rules? @FranchuFranchu

It is much simpler to write a performant interaction combinator runtime rather than a performant general interaction net runtime. When using interaction combinators, all nodes can use the same amount of memory, which simplifies allocation. It also means that nodes can be passed by value which removes indirection. We can even use a pre-existing interaction combinator runtime such as [HVM](https://github.com/HigherOrderCO/hvm) to reduce `par-lang` programs. The simplicity of interaction combinators means it's possible to quickly implement them in any platform, even GPUs for example. In addition, for linear logic, there is very little overhead to the translation to interaction combinators. Multiplicatives compile to single nodes and additives compile to 5-node trees. There aren't any huge complex interaction combinator nets. 

This is just daydreaming, but if in the future, an interaction net ecosystem springs up, we'd also be able to easily interface with other languages that use interaction combinators.

We could also compile to custom interaction nets, and that'd be good as a "default" runtime but perhaps won't be as good for a performant runtime. It would definitely be more performant than the current approach, though. However, if interaction combinators end up being strictly more performant than interaction nets, then I think it would be best to stick to interaction combinators.

There are already some runtimes to reduce general interaction nets, but they aren't as performant as the best interaction combinator runtimes.

PD: If you want to, we can communicate over Discord, which I believe is higher-throughput for conversations such as this one. But if you want to keep our conversation here, that's OK too.

faiface:
@FranchuFranchu Thanks for explaining why to go for interaction combinators. It's clearly the way to go. It would be great if the language could eventually run on something like HVM.

I haven't had too much time today, but I managed to put together a draft proposal for type syntax, which seems to be the logical next step: #6 

I'd love to hear your comments on it!

I also have an idea of a syntax for replicable types and values, which I'll draft tomorrow :) From what you're saying, it seems to be best to have the interaction combinator compiler before implementing replicable values, since they're going to be much easier to implement there compared to the current runtime.

About Discord, I'm all up for it, I'll set up a room tomorrow and add a link to the README, so that other people can join too, if they like.

I really appreciate your contributions so far!

PR #14, opened 2025-02-27
Unbreak playground run button

Easyoakland:
Fix run button.

### DIFF
--- a/src/playground.rs
+++ b/src/playground.rs
@@ -352,7 +352,8 @@ impl Playground {
 
                             ui.label(egui::RichText::new(error).color(red()).code());
                         }
-                    } else if let Some(int) = &self.interact {
+                    }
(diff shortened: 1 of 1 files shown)


faiface:
My bad, my goals was to hide the interaction area when compiled code is visible.

src/playground.rs
```diff
@@ -352,7 +352,8 @@ impl Playground {
 
                             ui.label(egui::RichText::new(error).color(red()).code());
                         }
-                    } else if let Some(int) = &self.interact {
+                    }
+                    if let Some(int) = &self.interact {
```
faiface, line 356: Could you please add a condition to only show when `!self.show_compiled`? That way will be like I originally intended.

faiface (changes requested):

PR #15, opened 2025-02-27
Example Folder

Easyoakland:
Move examples to example folder, and add the functional semigroup queue example.


faiface:
This is good to go, thanks!

PR #16, opened 2025-02-27
add scrollbar to run menu

Easyoakland:
makes the semigroup_queue.par example actually usable


faiface:
Neat! Thanks, merging

PR #22, opened 2025-03-09
Language Reference

s15n:
Currently WIP Language reference


docs/src/types.md
```diff
@@ -0,0 +1,273 @@
...
+```
+
+Mathematically, `!` is \\(\mathbf{1}\\), the unit for \\(\otimes\\).
+
+## The Bottom Type
+
+> **<sup>Syntax</sup>**\
+> _Bottom_ : `?`
+
+Bottom is an empty type. In Haskell, it's `void`, in Rust it's (perhaps confusingly) `!`.
```
faiface, line 262: This isn't true. `?` or _bottom_ is not the empty type. `either {}` or _0_ in LL is the empty type.

The bottom is actually the type of continuations with no input.

docs/src/items.md
```diff
@@ -0,0 +1,113 @@
...
+> &nbsp;&nbsp; &nbsp;&nbsp; _DefPattern_ (`,` _DefPattern_)<sup>\*</sup> `,`<sup>?</sup>
+> &nbsp;&nbsp; | `type` [_ID_List_]
+>
+> _DefPattern_ :\
+> &nbsp;&nbsp; &nbsp;&nbsp; [_PatternNoTopAlt_] `:` [_Type_]
+
+todo: patterns left of `=`
+-->
+
+`def` defines a replicable value usable throughout the file in which it was defined.
```
faiface, line 33: Not replicable, global definition. More in a comment below.

docs/src/items.md
```diff
@@ -0,0 +1,113 @@
...
+
+// define a function receiving types
+dec pop : [type T] [List<T>] (Option<T>) List<T>
+def pop(type T)(list: List<T>) = list {
+  .empty! => (.none!) .empty!
+  .item(head) tail => (.some(head)) tail
+}
+```
+-->
+Note: In reality, `greeting` has the [replicable]() type `&String`, while `negate` also has a replicable type: `&[Bool] Bool`.
```
faiface, line 76: Not true. `greeting` isn't replicable, it's just a global definition, which isn't a variable, but a global name that you can reference anywhere. Each time `greeting` is referenced, it creates a new instance of it, but it's not a single replicable value.

That's despite implementation details in the inet runtime, which sometimes turn globals into replicables.

The type `greeting` is `String` and the type of `negate is `[Bool] Bool`, no `&`. It's just that invoking their name creates a new instance because they are global definitions.

Linearity and "being replicable" only applies to local variables.

docs/src/expressions.md
```diff
@@ -0,0 +1,65 @@
...
+```
+
+\\[
+\frac{p:A^\perp \quad x:A \quad y\\{p/x\\}:B}{\texttt{let $p$ = $x$ in $y$}:B}
+\\]
+
+
+---
+
+## Tuple Expressions
```
faiface, line 41: Just like "tuple types" vs "pair types", I would opt for "pair expressions", and make it very clear that `(1, 2, 3)!` is a sugar for `(1) (2) (3) !`. It's important to realize that even in construction because this works just fine:

```
let pair = (2, 3)!
let triple = (1) pair   // triple = (1, 2, 3)!
```

docs/src/items.md
```diff
@@ -0,0 +1,113 @@
...
+}
+```
+<!--
+(or)
+```par
+// define a static value
+def greeting: String = "Hello, World!"
+
+// define a function
+def negate(b: Bool): Bool = b {
```
faiface, line 61: Not going to be a supported syntax

faiface (changes requested):
Left some comments. Great job, it's starting to look really good!

The syntax highlighting makes the language so pretty. I'd suggest also highlighting `!` and `?` because they are kinda special.

docs/src/items.md
```diff
@@ -0,0 +1,113 @@
...
+  .false! => .true!
+}
+// equivalent to
+def negate2: [Bool] Bool = [b] negate(b)
+
+// define a function receiving types
+dec pop : [type T] [List<T>] (Option<T>) List<T>
+def pop(type T)(list: List<T>) = list {
+  .empty! => (.none!) .empty!
+  .item(head) tail => (.some(head)) tail
```
Easyoakland: Shouldn't that be

`(.some(head)!)`
Or
`(.some head)`

docs/src/items.md
```diff
@@ -0,0 +1,113 @@
...
+def negate(b: Bool): Bool = b {
+  .true! => .false!
+  .false! => .true!
+}
+// equivalent to
+def negate2: [Bool] Bool = [b] negate(b)
+
+// define a function receiving types
+dec pop : [type T] [List<T>] (Option<T>) List<T>
+def pop(type T)(list: List<T>) = list {
```
Easyoakland: why `pop(type T)` instead of `pop[type T]`

docs/src/types.md
```diff
@@ -0,0 +1,416 @@
...
+  stack.push(.false!)
+  stack.pop
+} in stack
+```
+todo: Is this a good example?
+
+Mathematically, `{ .a => A, .b => B }` is \\(A \mathbin{\\&} B\\). For session types, it means "offer a choice of `A` or `B`".
+An empty choice `{}` is therefore \\(\top\\) and has exactly one value, `{}`. There is a function to it from every type:
+```par
+def terminate: [type T] [T] {} = [type T] [x] {}
```
Easyoakland, line 228: Is `x: T` unhandled, violating linearity?

docs/src/types.md
```diff
@@ -0,0 +1,416 @@
...
+```
+The result of this function can never be used though (todo: correct?).
```
Easyoakland, line 230: Why?

docs/src/types.md
```diff
@@ -270,4 +281,301 @@ let reverse = do {
...
+?def main = repeat(.succ.succ.succ.zero!, repl_bool(.true!))
+```-->
+
+Values of iterative types may be infinite (todo: correct?)
+```par
+type Inf<T> = iter (T) loop
+
+def infinite_bools: Inf<Bool> = begin (.true!) loop
+```
+This infinite value can be constructed but there is now way of fully destructing (so: using) it.
```
Easyoakland, line 407: Now -> no

docs/src/types.md
```diff
@@ -270,4 +281,301 @@ let reverse = do {
...
+```
+
+A more elaborate example can be seen [here](https://github.com/faiface/par-lang/blob/main/examples/flatten.par)
+
+A `chan A` can be linked with an `A`, annihilating both and ending the process.
+```par
+def just_true: Bool = chan yield {
+  let c: chan Bool = yield
+  let b: Bool = .true!
+  c <-> b
```
Easyoakland, line 542: Is it purposeful that you wrote `<->` instead of `<>`?

Easyoakland (commented):
Some notes

docs/src/items.md
faiface: It's not a supported syntax. The correct syntax is:

```
def pop = [type T] ...
```

But if it were supported syntax, then it would be like is written because things on the left of `=` get inverted, aka symmetric patterns.

docs/src/items.md
faiface: Yes! That should be `(.some head)` @s15n

docs/src/types.md
faiface: Nope, since `{}` can never ever be destroyed, `x` can live in it happily forever

faiface: But `terminate` is not a good name for this function @s15n

docs/src/types.md
faiface: Because there is no way to destroy a `{}`, the empty choice. It's the T type from LL, the unit for & ("with"). All you can do with choices is choose one of the variants. But there are no variants to choose here, hence nothing to do.

docs/src/types.md
s15n: I kind of thought so. What would be a good name?

docs/src/types.md
s15n: What do you mean?

docs/src/types.md
s15n: Looks like it's not finally decided whether it's <> or <->

docs/src/types.md
Easyoakland:
```diff
-This infinite value can be constructed but there is now way of fully destructing (so: using) it
+This infinite value can be constructed but there is no way of fully destructing (so: using) it
```

docs/src/types.md
```diff
@@ -0,0 +1,581 @@
...
+> _PairType_ : `(` _TypeList_ `)` _Type_
+
+Having multiple types between the `()` is just syntax sugar:
+```par
+type T = (A, B)!
+// is equivalent to
+type T = (A) (B) !
+```
+
+`!` is the unit for tuples, i.e. `(A, B)!` and `(A) B` are equivalent.
```
faiface, line 58: Saying "equivalent" here can be confusing because these types aren't equal. I know that's not what equivalent means, but I don't think it's necessary to emphasize these are equivalent. Also, "a unit for tuples" is a theoretical concept that's not really needed here.

Instead, I'd say that `(A, B)!` and `(A) B` are both valid ways to define a pair of `A` and `B` and depending on the context, one may be more convenient than the other.

docs/src/types.md
```diff
@@ -0,0 +1,581 @@
...
+// is equivalent to
+let x: either { .none!, .some T } = .none!
+```
+
+## Pair Types
+
+> **<sup>Syntax</sup>**\
+> _PairType_ : `(` _TypeList_ `)` _Type_
+
+Having multiple types between the `()` is just syntax sugar:
```
faiface, line 51: between `(` and `)`, or inside `(...)`

docs/src/types.md
```diff
@@ -0,0 +1,581 @@
...
+```
+
+Mathematically, `(A) B` is \\(A \otimes B\\). For session types, it means "send `A` and continue as `B`".
+
+## Function Types
+
+> **<sup>Syntax</sup>**\
+> _FunctionType_ : `[` _TypeList_ `]` _Type_
+
+Having multiple types between the `[]` is just syntax sugar:
```
faiface, line 94: Also, between `[` and `]`, or inside `[...]`.

docs/src/types.md
```diff
@@ -0,0 +1,581 @@
...
+```par
+type T = [A, B] R
+// is equivalent to
+type T = [A] [B] R
+```
+
+Values are created using [function expressions](./expressions.md#function-expressions):
+```par
+let id: [A] A = [a] a
```
faiface, line 103: I'd suggest putting a more interesting example here, perhaps in addition to the identity function.

docs/src/types.md
```diff
@@ -0,0 +1,581 @@
...
+```
+
+Mathematically, `[A] B` is a [linear](./linearity.md) function \\(A \multimap B\\). For session types, it means "receive `A` and continue as `B`".
+
+## Either Types
+
+> **<sup>Syntax</sup>**\
+> _EitherType_ : `either` `{` (_Label_ _Type_ `,`<sup>?</sup>)<sup>\*</sup> `}`
+
+An either type is a classical sum type aka. tagged union. Every value of such a type consists of a label (called "constructor") and a value of the type corresponding to the label (its "payload").
```
faiface, line 118: Wouldn't call it "classical" because classical refers to classical logic, which is relevant for Par here because of the distinction between intuitionistic and classical logic. A programmer not familiar with logic won't get confused here, but somebody familiar with logic actually might.

I'd just call it "the usual sum type"

faiface, line 118: The label isn't called "constructor", it's not a thing on its own. Like, `.variant` isn't a thing, only `.variant payload` is.

But, "consists of a label (called variant)", is completely valid.

docs/src/types.md
```diff
@@ -0,0 +1,581 @@
...
+}
+
+// a slightly more complex example
+type TwoOrNone<T> = either {
+  .none!      // constructor "none" with "no" payload (using !)
+  .two(T, T)! // constructor "some" with "two" payloads
+}
+```
+
+Values are created by calling one of the constructors with its payload.
```
faiface, line 134: Just a follow-up on the comment above, we're not "calling" constructors because they're not functions or any separate entity, `.variant payload` is the atomic thing.

docs/src/types.md
```diff
@@ -0,0 +1,581 @@
...
+
+> **<sup>Syntax</sup>**\
+> _ChoiceType_ :\
+> &nbsp;&nbsp; &nbsp;&nbsp; `{` (_Label_ (`(` _ReceiverList_ `)`)<sup>\*</sup> `=>` _Type_ `,`<sup>?</sup>)<sup>\*</sup> `}`
+>
+> _ReceiverList_ :\
+> &nbsp;&nbsp; &nbsp;&nbsp; _TypeList_ \
+> &nbsp;&nbsp; | `type` [_ID_List_]
+
+A choice type is dual to an [either](#either-types) type. Constructing a value of an either type is "making a choice" and similarly, destructing such a value looks exactly like constructing a value of a choice type.
```
faiface, line 162: "making a choice" is more of what the either-type does. I know what you meant, I think it's better expressed with "awaiting a choice", or something similar. Perhaps "offering multiple choices"

docs/src/types.md
```diff
@@ -0,0 +1,581 @@
...
+// define negate using the choice
+// featuring selecting from the choice type value
+def also_negate: [Bool] Bool = [b] b {
+  .true! => negate_choice.true
+  .false! => negate_choice.false
+}
+```
+`.cons => [A] B` can also be written as `.cons(A) => B`
+
+A choice type represents an interface for interacting with data. While an either type describes its underlying data, a choice type describes what can be done with it.
```
faiface, line 193: This is very well put!

docs/src/types.md
```diff
@@ -0,0 +1,581 @@
...
+}
+
+def main = do {
+  let list: List<Bool> = .item(.false!).empty!
+  let stack: Stack<List<Bool>, Bool>
+    = list_stack(type Bool)(list)
+
+  stack.push(.true!)
+  stack.push(.false!)
+  stack.pop
```
faiface, line 220: Just realized this `.pop` is missing something. You're not taking in the `Option<T>` it produces.

docs/src/types.md
```diff
@@ -0,0 +1,581 @@
...
+Mathematically, `{ .a => A, .b => B }` is \\(A \mathbin{\\&} B\\). For session types, it means "offer a choice of `A` or `B`".
+An empty choice `{}` is therefore \\(\top\\) and has exactly one value, `{}`. There is a function to it from every type:
+```par
+def terminate: [type T] [T] {} = [type T] [x] {}
+```
+The result of this function can never be used though (todo: correct?).
+
+Choice types are often used as [iterative](#iterative-types) types.
+
+## The Unit Type
```
faiface, line 234: Perhaps it would be a good idea to move this to the top of the list of types, as it's used in almost every other type somewhere

docs/src/types.md
```diff
@@ -0,0 +1,581 @@
...
+> **<sup>Syntax</sup>**\
+> _Bottom_ : `?`
+
+The bottom `?` is dual to the unit `!`. 
+It's mostly used in process syntax to destruct the unit:
+```par
+let triple: (A, B, C)! = ...
+
+let reverse = do {
+  triple[a, b, c]?
```
faiface, line 280: Hm, this is actually not an example of the `?` type, just the `?` operation. The `?` can construct the `?` type if it's used on the channel introduced by `chan`, like here:
```
def program = chan user {
  let continuation: ? = chan handover {
    handover?
    // do whatever with `user`
    user!
  }
  continuation!  // we can only end it with `!` in a process because it continues (thus ending this process)
}
```

docs/src/types.md
```diff
@@ -0,0 +1,581 @@
...
+} in (c, b, a)!
+```
+
+Mathematically, `?` is \\(\bot\\), the unit for \\(⅋\\). So \\(\bot \mathbin{⅋} A = \mathbf{1} \multimap A \cong A\\) (as seen before for the [unit](#the-unit-type) type).
+
+## Recursive Types
+
+> **<sup>Syntax</sup>**\
+> _RecursiveType_ :\
+> &nbsp;&nbsp; &nbsp;&nbsp; `rec` [_LoopLabel_]<sup>?</sup> _Type_
```
faiface, line 290: Gonna have to turn it to `recursive`

docs/src/types.md
```diff
@@ -0,0 +1,581 @@
...
+}
+```
+<!--// another way of defining a recursive type is the following:
+// Node is not recursive
+type Node<Next> = either {
+  .base!
+  .step Next
+}
+// Defining the recursive type
+type Rec = rec Node<self>
```
faiface, line 311: Yeah this actually works :D Nice

docs/src/types.md
```diff
@@ -0,0 +1,581 @@
...
+
+// We have the following subtyping relation:
+// Node<rec Node<self>> <: rec Node<self>-->
+
+Values of recursive types terminate (todo: correct?)
```
faiface, line 316: Correct, values of `recursive` types are always finite because they can only be constructed by a finite number of steps. They can't be constructed by a "formula". But, they can be destructed by a "formula": `begin`/`loop`.

On the other hand, `iterative` types can be constructed by a "formula", `begin`/`loop`, but cannot be destructed by a "formula", only using a finite number of destruction steps.

docs/src/types.md
```diff
@@ -0,0 +1,581 @@
...
+    .true! => !
+    .false! => !
+  }
+  .unwrap => b
+}
+?
+?def main = repeat(.succ.succ.succ.zero!, repl_bool(.true!))
+```-->
+
+Values of iterative types may be infinite (todo: correct?)
```
faiface, line 401: That's true, but only in the sense that they are defined by a "formula", a loop, which can be unrolled forever. The number of unrolls actualy performed will always be finite, though.

docs/src/types.md
```diff
@@ -0,0 +1,581 @@
...
+}
+```-->
+```par
+// construct a value of the iterative Repl<Bool>
+dec repl_bool : [Bool] Repl<Bool>
+// coinduction (marked by an independent begin-loop)
+def repl_bool = [b] begin {
+  // yield "(b, b)" and continue (coinductive step)
+  .copy => b {
+    .true! => (let b: Bool = .true! in loop, let b: Bool = .true! in loop)!
```
faiface, line 437: I would choose a simpler example here :D A stream of natural numbers, of fibonacci numbers will do a better justice because this example just gets kinda ugly.

docs/src/types.md
```diff
@@ -0,0 +1,581 @@
...
+> _ExistentialType_ : `(` `type` [_ID_List_] `)` _Type_
+
+Having multiple types between the `()` is just syntax sugar:
+```par
+type T = (type A, B) X
+// is equivalent to
+type T = (type A) (type B) X
+```
+
+Existential types mirror tuple types but they're qualified over types rather than values.
```
faiface, line 462: *pair types

docs/src/types.md
```diff
@@ -0,0 +1,581 @@
...
+They are used to obscure their underlying type.
```
faiface, line 463: "encapsulate" is probably better than "obscure"

docs/src/types.md
```diff
@@ -0,0 +1,581 @@
...
+?}
+```
+
+The qualifying types and values can both be extracted from a value of such a type:
+```par
+let any: Any = ...
+let (type X) x = any
+let y: X = x
+```
+todo: This doesn't work in Par currently. Is this intended?
```
faiface, line 488: It does work, what do you mean?

docs/src/types.md
```diff
@@ -0,0 +1,581 @@
...
+> _UniversalType_ : `[` `type` [_ID_List_] `]` _Type_
+
+Having multiple types between the `[]` is just syntax sugar:
+```par
+type T = [type A, B] X
+// is equivalent to
+type T = [type A] [type B] X
+```
+
+Existential types mirror function types but they're qualified over types rather than values.
```
faiface, line 504: *Universal types

And the description matches existential types more than universal types

docs/src/types.md
```diff
@@ -0,0 +1,581 @@
...
+type UnivEndo = [type T] [T] T
+type ParamEndo<T> = [T] T
+
+// a value of an universal type must be defined
+// for all types
+let id: UnivEndo = [type T] [x] x
+
+// a specialized version can be represented using
+// a parameterized type
+let id_bool: ParamEndo<Bool> = id(type Bool)
```
faiface, line 517: Some nice polymorphic function would be good here, for example some to reverse list ;) Or a less advanced one

docs/src/types.md
```diff
@@ -0,0 +1,581 @@
...
+```
+
+Mathematically, `[type T] A` is \\(\forall\ T: A\\).
+
+## Channel Types
+
+> **<sup>Syntax</sup>**\
+> _ChannelType_ : `chan` _Type_
+
+`chan A` represents a channel accepting an `A`:
```
faiface, line 527: Gotta mention that `chan T` turns `T` into its dual, ie `chan` is not a type constructor, but a type transformer.

docs/src/types.md
faiface: And we settled with `<>`

faiface (changes requested):
Got a new batch of review comments for you ;)

docs/src/types.md
s15n: 'Constructing a value of an either type is "making a choice"' would be correct, right?

docs/src/types.md
s15n: How would you solve this? (Currently it's being written as an output)

docs/src/types.md
faiface: Yes!

docs/src/introduction.md
```diff
@@ -0,0 +1,122 @@
...
+Par is centered around concurrency and session typing, all in the framework of linear logic.
+What does that mean?
+
+- Types are _linear_, i.e. a value must be used exactly once.
+  You might know the type system of Rust, where a value must be used at most once.
+- Ultimately, everything in Par is a _channel_.
+  - A list sends every item in order and then closes
+  - A function receives its argument and becomes the result
+  - An infinite stream can be signaled to either yield the next item or close
+- Channels communicate with each other by sending signals (the names with a dot in front)
```
faiface, line 39: That's just one of the ways channels communicate. Others are: sending values, and closing.

docs/src/introduction.md
```diff
@@ -0,0 +1,122 @@
...
+- Everything has a dual in Par. Every value is either supplied or consumed.
```
faiface, line 40: "either supplied or consumed" isn't accurate. It makes it sound like a whole value is supplied or consumed, but with LL, that becomes very granular and can change between supplying and consuming many times during the communication along a channel.

I would choose a different way of describing duality, perhaps by pointing out that each side of a channel sees the dual of the other?

faiface (changes requested):

docs/src/expressions.md
```diff
@@ -0,0 +1,57 @@
...
+dec apply_id : [type T] [T] T
+def apply_id = [type T] [x] {[y] y}(x)
+```
+
+## Let Expressions
+
+> **<sup>Syntax</sup>**\
+> _LetExpression_ : `let` [_Pattern_] `=` _Expression_ `in` _Expression_
+
+The second expression must use all bindings of the pattern (which must be irrefutable) due to linearity.
```
faiface, line 39: Would drop the "irrefutable", like we discussed.

Instead make it clear that:
- All patterns are currently irrefutable
- Patterns in matching are additionally prefixed by `.variant`, and must list all the variants of the type.

docs/src/expressions.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Do Expressions
+
+> **<sup>Syntax</sup>**\
+> _DoExpression_ : `do` `{` [_Process_] `}` `in` _Expression_
+
+Do expressions are syntax sugar for channel expressions (todo: specify how). Linearity requires that all leftover bindings from the process must be used in the expression at the end.
```
faiface, line 46: How is like this:

```
do {
  <commands>
} in <expression>

==

chan result {
  <commands>
  result <> <expression>
}
```

However, **all** expressions are syntax sugar on processes, so I don't think it's a good idea to highlight it here specifically.

I would instead describe `do` as a way to add commands to be executed before returning the expression after `in` because some things are better done with commands, and so it's nice to be able to insert them into expressions.

docs/src/expressions.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Channel Expressions
+
+> **<sup>Syntax</sup>**\
+> _ChanExpression_ : `chan` [ID] [_Annotation_]<sup>?</sup> `{` [_Process_] `}`
+
+The name declared after the channel may be used inside the process and must be fully destructed there.
```
faiface, line 53: must be used inside the process and fully destructed or moved out of there

docs/src/expressions.md
```diff
@@ -0,0 +1,57 @@
...
+The second expression must use all bindings of the pattern (which must be irrefutable) due to linearity.
+
+## Do Expressions
+
+> **<sup>Syntax</sup>**\
+> _DoExpression_ : `do` `{` [_Process_] `}` `in` _Expression_
+
+Do expressions are syntax sugar for channel expressions (todo: specify how). Linearity requires that all leftover bindings from the process must be used in the expression at the end.
+
+## Channel Expressions
```
faiface, line 48: Probably should expand more on this as it's the only kind of expression that can't be syntax sugar. It's super important for duality, and "construction by destruction", so will be key to understanding some patterns with processes.

docs/src/expressions/construction.md
```diff
@@ -0,0 +1,141 @@
+# Construction Expressions
```
faiface, line 1: Best to provide a single small example for each kind. Not practically useful ones (as that could obscure syntax), just ones that make the syntax clear.

docs/src/expressions/construction.md
```diff
@@ -0,0 +1,141 @@
...
+> _FunctionExpression_ : `[` [_PatternList_] `]` [_Expression_]
+
+*<sup>
+[Type](../types.md#function-types)
+| [Destructing Expression](application.md#function-calls)
+| [Statement](../statements/commands.md#todo)
+| [Destructing Statement](../statements/commands.md#todo)
+</sup>*
+
+If `p` is an irrefutable pattern for type `A` and `b` (wich must use the bindings of `p`) is of type `B`, the function expression `[p] b` is of the [function type](../types.md#function-types) `[A] B`.
```
faiface, line 58: Would also just say "pattern" here, and instead make it all clearer in the Patterns section.

docs/src/expressions/construction.md
```diff
@@ -0,0 +1,141 @@
...
+> &nbsp;&nbsp; &nbsp;&nbsp; `loop` [_LoopLabel_]<sup>?</sup>
+
+*<sup>
+[Type](../types.md#iterative-types)
+| [Statement](../statements/commands.md#todo)
+</sup>*
+
+If --- given every `loop` in `a` is of type `iterative A` --- `a` is of type `A`, the iterative construction `begin a` is of the [iterative type](../types.md#iterative-types) `iterative A`.
+
+If no loop label is present, `loop` corresponds to the innermost `begin`. Else to the `begin` with the same loop label.
```
faiface, line 112: Not true, a `loop` without a label only corresponds to a `begin` without a label. No label is a kind of a label.

docs/src/expressions/construction.md
```diff
@@ -0,0 +1,141 @@
...
+| [Statement](../statements/commands.md#todo)
+| [Destructing Statement](../statements/commands.md#todo)
+</sup>*
+
+If `a` is of type `A`, the existential construction `(type T) a` is of the [existential type](../types.md#existential-types) `(type T) A`.
+
+## Universal Constructions
+
+> **<sup>Syntax</sup>**\
+> _PairExpression_ : `[` `type` [_ID_List_] `]` [_Expression_]
```
faiface, line 132: Says PairExpression

docs/src/expressions/construction.md
```diff
@@ -0,0 +1,141 @@
...
+
+If --- given every `loop` in `a` is of type `iterative A` --- `a` is of type `A`, the iterative construction `begin a` is of the [iterative type](../types.md#iterative-types) `iterative A`.
+
+If no loop label is present, `loop` corresponds to the innermost `begin`. Else to the `begin` with the same loop label.
+
+
+## Existential Constructions
+
+> **<sup>Syntax</sup>**\
+> _PairExpression_ : `(` `type` [_TypeList_] `)` [_Expression_]
```
faiface, line 118: Says PairExpression

docs/src/expressions/application.md
```diff
@@ -0,0 +1,85 @@
+# Application Expressions
```
faiface, line 1: Examples demonstrating the syntax per application kind would be great. Not useful examples, but just to have the syntax in front of your eyes.

docs/src/expressions/application.md
```diff
@@ -0,0 +1,85 @@
...
+>
+> _LoopApplication_ :\
+> &nbsp;&nbsp; &nbsp;&nbsp; _Applicable_ `loop` [_LoopLabel_]<sup>?</sup>
+
+*<sup>
+[Destructs Type](../types.md#recursive-types)
+| [Statement](../statements/commands.md#todo)
+</sup>*
+
+If no loop label is present, `loop` corresponds to the innermost `begin`. Else to the `begin` with the same loop label.
```
faiface, line 71: Same as in constructions, not true, no label `loop` only corresponds to a no label `begin`.

docs/src/patterns.md
```diff
@@ -0,0 +1,268 @@
...
+> &nbsp;&nbsp; | [_PairPattern_](#pair-patterns) <!--\
+> &nbsp;&nbsp; | [_VariantPattern_](#variant-patterns) \
+> &nbsp;&nbsp; | [_BlankPattern_](#the-blank-pattern) -->\
+> &nbsp;&nbsp; | [_GroupedPattern_](#grouped-patterns) \
+> &nbsp;&nbsp; | [_ExistentialPattern_](#existential-patterns)
+>
+> _PatternList_ :\
+> &nbsp;&nbsp; &nbsp;&nbsp; _PatternNoAlt_ (`,` _PatternNoAlt_)<sup>*</sup> `,`<sup>?</sup>
+
+There are two properties of a pattern: What names it *binds* and whether it's *refutable*.
```
faiface, line 20: Let's drop the (ir)refutability for now, since it's not a concept, but instead let's explain that patterns currently are of form `x`, `x: T`, `(x) y`, and `!`. And that in matching, there is the `.variant` followed by a pattern. But with nuances, such as no `!` at the end of a pattern in choice construction.

docs/src/patterns.md
```diff
@@ -0,0 +1,268 @@
...
+The pattern `name` or `name: T` is always irrefutable and binds `name` to `x` when matching `x`.
+
+The pattern `name` can be used on a value which type is known. Otherwise, `name: T` must be used, which can only be used on a value of type `T`.
+
+```par
+let x = y in x
+// is just
+y
+
+// it is most often used for function expressions
```
faiface, line 95: Which part? The `lambda`, or the `x`? I wouldn't say "most often" in either case, patterns are fairly equally used for assigning variables (say before `begin`, or just intermediate steps), function arguments, and bindings from matchings.

docs/src/patterns.md
```diff
@@ -0,0 +1,268 @@
...
+let tuple: (A, B, C)! = (a, b, c)!
+tuple[first]
+// now first is a
+// and tuple is (b, c)!
+```
+
+If the type of the value matched is already known, a type annotation would have to match it.
+Such an annotation can be useful for declaring functions, though:
+```par
+// no extra type annotation/declaration needed
```
faiface, line 108: Unfortunately, this wouldn't type-check because the type fo the result isn't specified and it returns an either: not inferrable.

docs/src/patterns.md
```diff
@@ -0,0 +1,268 @@
...
+```par
+dec get_second : [type T] [List<T>] (Option<T>) List<T>
+def get_second = [type T] [list] list {
+  .item(fst).item(snd) rest => (.some snd) .item(fst) rest,
+  other => (.none!) other
+}
+```
+-->
+
+## Grouped Patterns
```
faiface, line 213: Grouped patterns don't exist

faiface (changes requested):
Reviewed expressions and patterns, still pending to review statements.

docs/src/statements/commands.md
```diff
@@ -0,0 +1,525 @@
...
+> _LoopLabel_ :\
+> &nbsp;&nbsp; &nbsp;&nbsp; `:` [ID]
+
+*<sup>
+[Destructs Type](../types.md#recursive-types)
+| [Destructs Channel](../types.md#iterative-types)
+| [Expression](../expressions/application.md#recursive-destructions)
+</sup>*
+
+If no loop label is present, `loop` corresponds to the innermost `begin`. Else to the `begin` with the same loop label.
```
faiface, line 329: Just the same as before, no label is a label.

docs/src/statements/commands.md
```diff
@@ -0,0 +1,525 @@
...
+    // either { .empty!, .item(Bool) List<Bool> }
+    // notice the absence of recursive
+    .item => {
+      // list is now of type (Bool) List<Bool>
+      list[b]
+      b {
+        .true! => {}
+        .false! => {
+          drop_bool(result)?
+          let result: Bool = .false!
```
faiface, line 348: Would be good to mention how context variables work across loops. That variables at `loop` must be the same and equal types as at `begin`.

It may also be confusing to some, that we're re-assiging `result` here, as a new variable, but it's carried across the loop. That's probably a good idea to point out, that variables can be re-assigned, as new variables, and it's okay even across loops.

docs/src/statements/commands.md
```diff
@@ -0,0 +1,525 @@
...
+  // stream is now again Stream<T>
+
+  // combine both operations
+  stream.next[second]
+
+  // close the stream
+  stream.close?
+} in (first, second)!
+```
+It can also be used in channels: Destructing a choice type to construct an either type:
```
faiface, line 209: "in channels", I get what you're trying to say here. The channels introduced at `chan`. But that's a little inaccurate because all variables are channels.

I think a better wording would be: It can also be used with `chan` to dually construct ...

And apply this to all the other usages of "used in channels".

docs/src/statements/commands.md
```diff
@@ -0,0 +1,525 @@
...
+could have been replaced with
+```par
+.item(b) => {
+  ...
+}
+```
+
+A recursive command can also be used in channels: Destructing a recursive (here: either) type to construct an iterative (here: choice) type:
+```par
+// chan Stream<Bool> is
```
faiface, line 375: The examples are amazing! But I'd try to make some of them shorter. This long they can be difficult to read. I'd try to condense them so that the idea relevant to the discussed command shines through.

faiface (changes requested):
Some comments to the Statements section!

Gotta say, I really love the section, it's very well put together.

docs/src/expressions/construction.md
```diff
@@ -215,9 +316,22 @@ dual <> a
 | [Destructing Statement](../statements/commands.md#send-type-commands)
 </sup>*
 
+Having multiple names between `(` and `)` is just syntax sugar:
```
Easyoakland, line 319: Either `()` is wrong or the below example is the wrong example.

docs/src/expressions/construction.md
```diff
@@ -215,9 +316,22 @@ dual <> a
...
+```par
+[type T, U] x
+// is equivalent to
+[type T] [type U] x
+```
+
 If `a` is of type `A`, the universal construction `[type T] a` (where `a` can use the type `T`) is of the [universal type](../types.md#universal-types) `[type T] A`.
 
-todo
+Universal constructions are moslty used to define "generic functions":
```
Easyoakland, line 328: Typo "moslty"

faiface:
This is wonderful! I'm going to merge and set it up now, any pending corrections to be resolved after the merge.

PR #26, opened 2025-03-14
Combinator parser

Easyoakland:
Unfinished version of porting the parser to combinators.

The advantage with this should be
1. Error messages can be customized.
2. A completely custom parser can iteratively replace this since its just Rust functions.
3. [`miette`](https://crates.io/crates/miette) should be able to easily integrate with this to make the errors very pretty

While it's still unfinished let me know if it looks unexpectedly good/bad.


Easyoakland:
It should now be possible to compare the old parser vs the new to see if the errors are better.

Both are reported on syntax errors. Let me know if there's something that should be improved/changed.

src/par/types.rs
```diff
@@ -2188,92 +2224,106 @@ impl<Loc, Name: Display> TypeError<Loc, Name> {
...
+                    "Operations cannot be performed on the same type."
                 )
             }
             Self::NoSuchLoopPoint(loc, _) => {
-                format!(
-                    "{}\nThere is no matching loop point in scope.",
-                    display_loc(loc),
-                )
+                let labels = labels_from_loc(code, loc);
+                miette::miette!(labels = labels, "There is no source_matching loop point in scope.")
```
faiface, line 2283: source_matching -> matching

faiface (changes requested):
So good!

I added one comment about a typo. Aside from that, let's remove the old syntax errors display for comparison.

After that, let's merge this!

Easyoakland:
Do we keep the now unnecessary `syntax.pest`? I would say no since the language reference covers the same content and the parser is already diverging from the `syntax.pest` such as the handling of loops, matches, and unfounded.

faiface:
@Easyoakland Nah, we don't need to keep it.

Not sure if you noticed, but one more change I said was to remove the errors for comparison. You're now still displaying two versions of syntax errors to compare.

Easyoakland:
Working on that now.

Easyoakland:
Alright. Pest parser is gone and the new parser is in the same place it used to be.

faiface (approved):

PR #30, opened 2025-03-28
run `dist init` and add repo url to Cargo.toml

Easyoakland:
Ran `dist init` and selected default settings.

This should build new releases when you push a tag which looks like a release e.g. `0.1.2`

See https://github.com/axodotdev/cargo-dist and https://opensource.axo.dev/cargo-dist/book/introduction.html

If you would rather manually generate releases that can be changed by changing [this setting](https://opensource.axo.dev/cargo-dist/book/reference/config.html?highlight=workflow%20dispatch#dispatch-releases) in the `dist-workspace.toml` and running `dist init` again.


Easyoakland:
Also, added a `CHANGELOG.md` file where the which `dist` uses for release notes, and is useful in and of itself.

PR #31, opened 2025-03-28
Language server

s15n:
For this we need:

- [ ] Fully typed AST after compilation 
- [ ] every one of its Name nodes stores the declaration/definition and **all references**

### DIFF
--- a/src/par/lexer.rs
+++ b/src/par/lexer.rs
@@ -11,12 +11,12 @@ use winnow::{
 
 #[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
 pub enum TokenKind {
-    LPar,
-    RPar,
-    LCrl,
-    RCrl,
-    LBkt,
-    RBkt,
+    LParen,
+    RParen,
+    LCurly,
+    RCurly,
+    LBrack,
+    RBrack,
     Lt,
     Gt,
 
@@ -70,12 +70,12 @@ impl PartialEq<TokenKind> for Token<'_> {
 impl TokenKind {
     pub fn expected(&self) -> &'static str {
         match self {
-            TokenKind::LPar => "(",
-            TokenKind::RPar => ")",
-            TokenKind::LCrl => "{",
-            TokenKind::RCrl => "}",
-            TokenKind::LBkt => "[",
-            TokenKind::RBkt => "]",
+            TokenKind::LParen => "(",
+            TokenKind::RParen => ")",
+            TokenKind::LCurly => "{",
+            TokenKind::RCurly => "}",
+            TokenKind::LBrack => "[",
+            TokenKind::RBrack => "]",
             TokenKind::Lt => "<",
             TokenKind::Gt => ">",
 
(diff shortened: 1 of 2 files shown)


src/par/lexer.rs
```diff
@@ -10,31 +11,48 @@ use winnow::{
...
 pub enum TokenKind {
-    LParen,
-    RParen,
-    LCurly,
-    RCurly,
-    LAngle,
-    RAngle,
-    LBrack,
-    RBrack,
+    LPar,
```
faiface, line 14: Actually I think the previous names were better here, more obvious on the first look.

src/par/parse.rs
```diff
@@ -351,41 +305,42 @@ impl ProgramParseError {
...
+
 fn program(
     mut input: Input,
-) -> std::result::Result<Program<Loc, Name, Expression<Loc, Name>>, ProgramParseError> {
-    enum Either<A, B, C> {
-        A(A),
-        B(B),
-        C(C),
+) -> std::result::Result<Program<Name, Expression<Name>>, ProgramParseError> {
+    pub enum Item<Name, Expr> {
```
faiface, line 312: Very good!

src/par/types.rs
```diff
@@ -7,91 +7,103 @@ use std::{
...
-    Break(Loc),
-    Continue(Loc),
-    Recursive(Loc, IndexSet<Option<Name>>, Option<Name>, Box<Self>),
-    Iterative(Loc, IndexSet<Option<Name>>, Option<Name>, Box<Self>),
-    Self_(Loc, Option<Name>),
-    SendType(Loc, Name, Box<Self>),
-    ReceiveType(Loc, Name, Box<Self>),
+pub enum Type<Name> {
+    Chan(Span, Box<Self>),
+    // todo: difference?
```
faiface, line 65: It's for specifying that the type is certainly a type variable and cannot be expanded. A `Type::Name` can refer to a type variable, but then if expanded, expandes to `Type::Var`, which is no longer expanded.

src/par/types.rs
```diff
@@ -7,91 +7,103 @@ use std::{
...
+    Receive(Span, Vec<Self>, Box<Self>),
+    Either(Span, IndexMap<Name, Self>),
+    Choice(Span, IndexMap<Name, Self>),
+    /// ! (unit)
+    Break(Span),
+    /// ? (bottom)
+    Continue(Span),
+    Recursive {
+        span: Span,
+        // todo: what is this?
```
faiface, line 78: Not the most descriptive name, I admit :D It's the ascendents of the type (denoted by the names of the respective loop points), meaning, if you `begin` on a `recursive`, and it expands, so its `self`s get replaced by new `recursive`s, these new `recursives` will have as their ascendent the original `recursive`. This is for totality checking.

src/main.rs
```diff
@@ -21,4 +28,72 @@ async fn main() {
         Box::new(|cc| Ok(Playground::new(cc))),
     )
     .expect("egui crashed");
+}*/
+
+
+fn main() {
+    language_server::language_server_main::main();
```
faiface, line 35: Gotta add a way to start a language server, or start a playground. Some command line flag.

faiface (changes requested):
Amazing job! The refactoring too.

Too changes:
- Just a nitpick, but I really think the previous bracket names in the lexer were better
- Gotta add a way to both start the playground, and the language server.

faiface (approved):
Great, merging!

PR #34, opened 2025-04-02
fix semigroup queue example for totality changes

Easyoakland:
Removes the recurse-by-name instances in the `semigroup_queue.par` example so it works again.


faiface:
Wonderful, thanks!

PR #40, opened 2025-04-18
Compact type annotations, name: type on hover on names in LSP

s15n:
Changed up the type of one function a bit, might need to be renamed or reverted


faiface:
Very nice! I'm enjoying the types in VSCode a lot :D Just one thing: cargo fmt

Then I'm merging.

faiface (changes requested):
cargo fmt

faiface (approved):

PR #41, opened 2025-04-21
Rw counter

faiface:


faiface:
Was hard to merge, so I instead copied the functionality from the commits

Issue #9, opened 2025-02-07
Improvements to the playground


faiface:
There's a bunch of features that would highly benefit the current playground. Let's keep a checklist here:

- [ ] Open/save files.
- [ ] Auto-compile on Run, if needed.
- [ ] Single-click to run the most recently executed definition.
- [ ] Scroll-bars of the "compiled code" editor and the whole interactive area are overlapping, would be nice to fix.
- [ ] Code theme selection.
- [ ] Dark/light/system switch with the first two options overriding system dark/light mode.
- [ ] Web version.

Comment down below if you want to give one of these a go, or if you suggest another (ideally simple) missing feature.

notJoon:
I have found that when non-ASCII UTF-8 characters are entered into the playground, the font appears broken (a.k.a tofu). it would be good to modify it so that these characters are rendered properly as well.

NoamDev:
Code theme selection is actually complete;)

faiface:
Most of these remaining features, perhaps except for the code theme selection, don't seem like they would bring much, so closing this issue.

PR #42, opened 2025-04-28
feat: Add theme mode selection

notJoon:
# Description

- Add theme mode selection (System/Light/Dark) with a dropdown menu
- consistent theme across all UI components

Related with #9 

## Example

**Light Mode**

Screenshot 2025-04-28 at 6 59 04 PM

**Dark Mode**

Screenshot 2025-04-28 at 6 59 48 PM

### DIFF
--- a/src/playground.rs
+++ b/src/playground.rs
@@ -35,6 +35,42 @@ pub struct Playground {
     show_ic: bool,
     readback_state: Option<crate::readback::ReadbackState>,
     cursor_pos: (usize, usize),
+    theme_mode: ThemeMode,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ThemeMode {
+    System,
+    Light,
+    Dark,
+}
+
+impl ThemeMode {
+    pub fn is_dark(&self, system_dark: bool) -> bool {
+        match self {
+            Self::System => system_dark,
+            Self::Dark => true,
+            Self::Light => false,
+        }
+    }
+
+    pub fn display_name(&self) -> &'static str {
+        match self {
+            Self::System => "System",
+            Self::Light => "Light",
+            Self::Dark => "Dark",
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


faiface (changes requested):
Wonderful!

There is one problem, though. The System theme doesn't work right.

When I switch to Light, and then to System, it stays light. If I switch to Dark, and then System, it stays dark.

In other words, System seems to use the last selected theme.

notJoon:
@faiface Thank you for your feedback. I have fixed the issue in this [commit](https://github.com/faiface/par-lang/pull/42/commits/f9703242df969fbcc31853dbb2d94ca57227d843). Could you please check it? I've added a demo video to the description for your reference. Thank you 👍

faiface (changes requested):
Works great! Nice video too ;)

One more small thing:

Untitled

Somehow we got in these grey backgrounds on some labels, not sure how that happened...

Before your change, it was like this:

Screenshot 2025-04-28 at 14 54 10

Could you try and fix this? After that, it's perfect.

notJoon:
> Could you try and fix this? After that, it's perfect.

After checking, I found that `set_visual` was being called on every frame, which was overriding all styles with default values. So the custom styles were also being reset, causing the background color to appear unintentionally. I've modified it to reapply the custom styles after each `set_visuals` call on every frame.

Screenshot 2025-04-28 at 10 50 51 PM

After the modification, it appears that the issue has been resolved.

notJoon:
> Working with UI is fun, isn't it? 😄
@faiface 
Yep The UI is certainly quite fun (honestly, I didn't expect it to be this difficult 😅). I applied your suggestions and did more testing to confirmed the following points:

- Background color remains transparent for all labels except selectable buttons
- Even when the system theme changes, the theme explicitly selected by the user is maintained

Below is the demo

Screen Recording 2025-04-29 at 10 36 10 AM

Also, `cargo fmt` too

PR #45, opened 2025-04-29
fix: playground crash when compiles programs with only type definitions

### LINKED ISSUE #43: Crash when no `def` exist
The following program crashes the playground:
```
type Either = either {.a ! .b !}
```
and so does an empty program.

This does not
```
type Either = either {.a ! .b !}
def a = !
```

the crash is here
https://github.com/faiface/par-lang/blob/655cc890451a6d97357516a9830989ff7b35b055/src/par/program.rs#L177

notJoon:
# Description

The playground was crashing when trying to compile a programs containing only type definitions (like `type Either = either {.a ! .b !}`).

This was happening because the `TypeOnHover::query` method was attempting to access an empty `sorted_pairs` vector.

The fix adds a check for empty `sorted_pairs` at the beginning of the query method, returning `None` immediately if there are no type definitions to query. This ensures the playground handles these cases without crashing.

closes #43

### DIFF
--- a/src/par/program.rs
+++ b/src/par/program.rs
@@ -162,9 +162,9 @@ impl<Name: Clone + Spanning> TypeOnHover<Name> {
 
 impl<Name: Clone> TypeOnHover<Name> {
     pub fn query(&self, row: usize, column: usize) -> Option<NameWithType<Name>> {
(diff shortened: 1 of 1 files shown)


faiface:
Thank you very much @notJoon! Merging

PR #54, opened 2025-06-24
use `commit_after` when parsing `<...>`

Easyoakland:
Remove a TODO in the Parser regarding parsing of `<...`>.

I think I forgot to remove the TODO when re-writing the parser implementation.

Effectively makes it so that
```
type Stream<T>
```
Errors because `T` is uppercase instead of highlighting all of `<T>` with a more ambiguous error.


faiface (approved):
Very nice! Thanks a lot!

PR #56, opened 2025-06-27
Added Int.FromString (And made CLI errors prettier)

NoamDev:


faiface (changes requested):
Nice! I left some comments, happy to merge after fixing.

Also one optional one: what about adding `Nat.FromString` as well?

faiface (approved):
Great, looks good now!

PR #58, opened 2025-07-08
Fix spelling and grammar mistakes in documentation.

FranchuFranchu:
Also added `cargo-spell` configuration.


docs/src/structure/definitions_and_declarations.md
```diff
@@ -30,7 +30,7 @@ remedied by totality checkers, such as in [Agda](https://en.wikipedia.org/wiki/A
 Par chooses a different approach. That is outlawing unrestricted recursion, and instead relying on more
 _principled_ ways to achieve cyclic behavior, including what's usually achieved by mutual recursion.
 
-**The remedies** come in form of these more principled ways. Fortunately, they don't just replace the
+**The remedies** come the the form of these more principled ways. Fortunately, they don't just replace the
```
faiface, line 33: the the

faiface (changes requested):
Very good, thank you so much!

Just found one new typo in your PR ;)

And I have one question: in what situations will I need to be updating the configuration for the spellcheck, going forward?

FranchuFranchu:
> And I have one question: in what situations will I need to be updating the configuration for the spellcheck, going forward?

@faiface , whenever you use vocabulary that isn't in `cargo-spellcheck`'s builtin English dictionary, you'll have to add an entry into the `.config/lingo.dic` file. It is basically a list of words, but it supports "tags" which specify what affixes can be added into the words (for example, "desugar" can be turned into "desugared", hence the `/D` tag). Here's the builtin English dictionary, which can be useful to see which words have which tags [https://github.com/drahnr/cargo-spellcheck/blob/master/hunspell-data/en_US.dic](https://github.com/drahnr/cargo-spellcheck/blob/master/hunspell-data/en_US.dic).

faiface:
@FranchuFranchu I see, thanks for explaining! This is a good feature, I'm merging your PR :)

PR #64, opened 2025-07-19
feat: Backup playground contents and print to stderr on panic.

Easyoakland:
The recent demo motivated me to finally fix the worry about losing everything when you forget to save and playground crashes.

Now crash will print code to stderr before exit.

Clones the string for the entire playground file every render, but testing indicated it wasn't too slow at 50,000 lines.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -7,6 +7,7 @@ use colored::Colorize;
 use eframe::egui;
 use futures::task::SpawnExt;
 use std::fs::File;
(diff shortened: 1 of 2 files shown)


Easyoakland:
You can add
```rust
if ui.button("panic!").clicked() {
  panic!("panic! button clicked")
}
```
if you want to test it.

faiface (approved):
This is a great idea! Thanks a lot, merging immediately

PR #66, opened 2025-07-24
feat: unit test

notJoon:
# Description

This PR implements a test framework for the Par, adding a new `Test.Test` type that support assertions per test with descriptive messages.

### Core Components

1. **AssertionResult** (`src/test_assertion.rs`)
   - Stores assertion description and pass/fail status

2. **Test Provider** (`src/par/builtin/test.rs::provide_test`)
   - Implements the Test type behavior
   - Handles assert method calls
   - Sends results through mpsc channel

3. **Test Runner** (`src/test_runner.rs`)
   - Creates mpsc channel for each test
   - Injects Test instance into test functions
   - Collects and reports assertion results

### Test Discovery
1. Test runner scans for functions starting with "Test" (excluding "Test" itself)
2. Validates that test functions have signature `[Test.Test] !`

### Test Execution Flow
1. **Channel Creation**: Create mpsc sender/receiver pair for assertion collection
2. **Test Instance Creation**: 
   ```rust
   // PATH: src/test_runner.rs
   let untyped_handle = test_handle.into_untyped();
   provide_test(untyped_handle, sender_clone);
   ```
3. **Test Execution**: Test function receives Test instance and calls assertions
4. **Result Collection**: Assertions are sent through mpsc channel
5. **Reporting**: Failed assertions are displayed with descriptions

## Test Example

```par
def TestAddCommutative: [Test.Test] ! = [test] do {
  test.assert("Add should be commutative", Nat.Equals(Add(2, 3), Add(3, 2)))
} in !
```

### Output
스크린샷 2025-07-24 오후 10 24 28


examples/TestExample.par
```diff
@@ -0,0 +1,27 @@
+def Add: [Nat, Nat] Nat = [a, b]
+  Nat.Add(a, b)
+
+def Mul: [Nat, Nat] Nat = [a, b]
+  Nat.Mul(a, b)
+
+def TestAddCommutative: [Test.Test] ! = [test] do {
```
faiface, line 7: The way global name resolution is done, `Test` and `Test.Test` are treated the same. Ie, `Anything.Anything` is always the same as `Anything`. So `[Test] !` is sufficient.

examples/fibonacci.par
```diff
@@ -37,3 +37,35 @@ def Program: ! = do {
   fib.close
   console.close
 } in !
+
+def TestNthFibZero: [Test.Test] ! = [test] do {
+  test.assert("NthFib(0) should be 1", Nat.Equals(NthFib(0), 1))
+} in !
+
+def TestNthFibOne: [Test.Test] ! = [test] do {
```
faiface, line 45: Let's combine this into a single test, like:
```
def TestNthFib: [Test] ! = [test] do {
  let results: List<Nat> = *(1, 1, 2, 3, 5, 8, 13, 21)
  let index: Nat = 0
  results.begin.case {
    .end! => {}
    .item(correct) => {
      let message = String.Builder
        .add("NthFib(")
        .add(Nat.ToString(index))
        .add(") should be ")
        .add(Nat.ToString(correct))
      test.assert(message, Nat.Equals(NthFib(index), correct))
      let index = Nat.Add(index, 1)
    }
  }
} in !
```

src/icombs/readback.rs
```diff
@@ -667,3 +678,45 @@ pub fn expand_type(typ: Type, type_defs: &TypeDefs) -> Type {
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::icombs::compiler::TypedTree;
+    use crate::location::Span;
+    use crate::par::types::Type;
+    use std::sync::atomic::AtomicUsize;
+
+    #[test]
+    fn test_typed_handle_into_untyped() {
```
faiface, line 691: I would skip this test. It's true we have very few tests in the codebase, which could be improved. But this test only tests a basic functionality that can be seen to be correct on the first look, so its added value is questionable.

src/par/builtin/test.rs
```diff
@@ -0,0 +1,72 @@
...
+    test_assertion::AssertionResult,
+};
+
+pub fn external_module() -> Module<Arc<process::Expression<()>>> {
+    Module {
+        type_defs: vec![TypeDef {
+            span: Span::None,
+            name: GlobalName::external(Some("Test"), "Test"),
+            params: vec![],
+            typ: Type::iterative_box_choice(vec![(
```
faiface, line 21: Is
```
Type::iterative(None, Type::box_(Type::choice(vec![
```
really that bad? I think it reads okay and not worth adding a whole constructor function to `Type`.

src/par/builtin/test.rs
```diff
@@ -0,0 +1,72 @@
...
+                    let passed = match bool_handle.case().await.as_str() {
+                        "true" => {
+                            bool_handle.break_();
+                            true
+                        }
+                        "false" => {
+                            bool_handle.break_();
+                            false
+                        }
+                        variant => {
```
faiface, line 50: Just like in other builtin definitions
```
_ => unreachable!(),
```
is completely sufficient here. A panic there will give you a line number, and the situation just cannot occur unless:
- Types are specified wrong
- The external impl isn't going according to the types
- There is a bug in the type checker

In any case, the line number on `unreachable!()` gives a sufficient diagnostic for this.

src/par/builtin/test.rs
```diff
@@ -0,0 +1,72 @@
...
+                        }
+                    };
+
+                    // Send the assertion result
+                    let result = AssertionResult {
+                        description,
+                        passed,
+                    };
+
+                    let _ = sender.send(result);
```
faiface, line 64: Checking the error here is a good idea.
```
sender.send(result).expect("receiver dropped");
```
will do the trick. The receiver dropped indicates a bug, gonna be much easier to debug if it crashes here, as opposed to having to notice you don't see everything printed.

src/par/types.rs
```diff
@@ -226,6 +226,29 @@ impl Type {
...
+                result.fold(last, |acc, t| Self::pair(t, acc))
+            }
+        }
+    }
+
+    pub fn iterative_box_choice(branches: Vec<(&'static str, Self)>) -> Self {
+        // FIXME: Proper implementation of iterative box choice is needed
+        // For now, just use a regular box choice
+        // This means tests can only make one assertion
+        Self::box_(Self::choice(branches))
```
faiface, line 250: Ohhh :D Now it's even misleading :D Let me know what help you need to implement an iterative box choice!

And yeah, I'd remove this constructor.
```
Type::iterative(None, Type::box_(Type::choice(vec![
```
is completely sufficient.

src/par/types.rs
```diff
@@ -226,6 +226,29 @@ impl Type {
             }),
         )
     }
+
+    pub fn bool() -> Self {
```
faiface, line 230: I'm on the fence on this one, but I'm still leaning to this definition being redundant. Other code uses
```
Type::name(Some("Bool"), "Bool", vec![])
```
for bools (because they're globally defined), and the above constructor makes it seems like bools are a primitive, which is misleading.

src/par/types.rs
```diff
@@ -226,6 +226,29 @@ impl Type {
...
+        Self::either(vec![("false", Self::break_()), ("true", Self::break_())])
+    }
+
+    pub fn tuple(types: Vec<Self>) -> Self {
```
faiface, line 234: This suffers even more from what I wrote above, which is it makes it seem like tuples are some kind of a base language type, which they are not. Furthermore, the constructor is smart, as in for length 0 it produces `!`, for length one it produces `T` and for N>1 it produces `(T1, T2, ...) TN`.

I think seeing it used can end up more confusing than helpful.

src/test_assertion.rs
```diff
@@ -0,0 +1,75 @@
...
+
+        let result2 = AssertionResult {
+            description: "Second test".to_string(),
+            passed: false,
+        };
+
+        sender.send(result1.clone()).unwrap();
+        sender.send(result2.clone()).unwrap();
+
+        assert_eq!(receiver.recv().unwrap(), result1);
```
faiface, line 47: Since the definition of `create_assertion_channel()` is `mpsc:channel()`, this is testing that mpsc channel works, which we don't have to test. As such, I'd remove these tests too.

src/test_runner.rs
```diff
@@ -0,0 +1,395 @@
...
+    let type_defs = program.type_defs.clone();
+    let spawner = Arc::new(TokioSpawn::new());
+    let sender_clone = sender.clone();
+
+    // Spawn the test function execution
+    let test_future = spawner
+        .spawn_with_handle(async move {
+            let handle = TypedHandle::from_wrapper(type_defs, net_wrapper, tree);
+
+            // The test function expects [Test.Test] !
```
faiface, line 205: You need to verify that this is true and not run the test if it isn't. You can check it with `Type::is_assignable_to`.

src/test_runner.rs
```diff
@@ -0,0 +1,395 @@
...
+            duration.as_secs_f32()
+        )
+        .red()
+    };
+
+    println!("{}", summary);
+}
+
+#[cfg(test)]
+mod tests {
```
faiface, line 324: These tests also test very little. Would remove them too.

src/par/builtin/test.rs
```diff
@@ -0,0 +1,72 @@
...
+        declarations: vec![],
+        definitions: vec![],
+    }
+}
+
+pub fn provide_test(handle: Handle, sender: mpsc::Sender<AssertionResult>) {
+    handle.provide_box(move |mut handle| {
+        let sender = sender.clone();
+        async move {
+            // for iterative box choice, we need to handle each method call
```
faiface, line 35: Yep, for that you just need to loop.

faiface (changes requested):
Thanks and sorry for my delayed review. This is very much on the right track, we need to polish some things up:

- The `#[test]`s I found were mostly without much added value, I'd remove them, so we don't have to maintain them. It's true that tests in the repo are generally lacking, but to fix that, we need to think where tests would provide most benefit and add them in such a way that doesn't require too much overhead. These tests here generally don't do either.
- Gotta make it `iterative box choice`, to have that, you just need to loop in the external implementation. Perhaps take inspiration from `String.Builder` external impl.
- A couple of other changes requested in the review comments.

PR #73, opened 2025-07-31
fix loop inference in case

### LINKED ISSUE #69: Bug: Inference fails for process syntax with loops
For example, in `examples/Fibonacci.par`, the main program is specified as:

```
def Program: ! = do {
  let console = Console.Open
  let fib = Fibonacci

  Nat.Repeat(30).begin.case {
    .end! => {}
    .step remaining => {
      fib.next[n]
      console.print(Nat.ToString(n))
      remaining.loop
    }
  }

  fib.close
  console.close
} in !
```

Notice the `Program: !` annotation. That's because currently, the inference fails if we remove the annotation:

```
def Program = do {
  let console = Console.Open
  let fib = Fibonacci

  Nat.Repeat(30).begin.case {
    .end! => {}
    .step remaining => {
      fib.next[n]
      console.print(Nat.ToString(n))
      remaining.loop
    }
  }

  fib.close
  console.close
} in !
```

Here we get:

```
  × Operations cannot be performed on the same type.
    ╭─[33:16]
 32 │       console.print(Nat.ToString(n))
 33 │       remaining.loop
    ·                ──┬──
    ·                  ╰── should operate on the same type as this
 34 │     }
 35 │   }
 36 │ 
 37 │   fib.close
 38 │   console.close
 39 │ } in !
    ·      ┬
    ·      ╰── this
    ╰────
```

We know that the inference failure is related to loops because if we comment out the loop line, it works:

```
def Program = do {
  let console = Console.Open
  let fib = Fibonacci

  Nat.Repeat(30).begin.case {
    .end! => {}
    .step remaining => {
      fib.next[n]
      console.print(Nat.ToString(n))
      //remaining.loop
    }
  }

  fib.close
  console.close
} in !
```

Something is wrong with the combination of loops and branch unification. This shouldn't be a big problem to fix.

NoamDev:
fixes #69

### DIFF
--- a/src/par/types.rs
+++ b/src/par/types.rs
@@ -1953,6 +1953,9 @@ impl Context {
                     };
                     self.put(span, object.clone(), branch_type.clone())?;
                     let (process, inferred_in_branch) = analyze_process(self, process)?;
+                    if matches!(inferred_in_branch, Some(Type::Self_(_,_))) {
+                        inferred_type = None
+                    }
                     typed_processes.push(process);
 
                     match (inferred_type, inferred_in_branch) {


NoamDev:
I wonder what should be the solution then.
Introducing a new specialized type would be too much, right?
Should we make the infer commands return something more complicated?

faiface:
I guess it might make sense to unify "self" and "A" as "A".

But we need to reintroduce unification for the inferred types in branches in `.case`.

Right now there is that subtype check which isn't even fully correct. Unification is what's needed.

It's implemented back in history if you scroll in commit history far enough, I made the mistake of removing it :D

NoamDev:
Yeah it seems we must resolve this the correct way. Where a loop that isn't on the subject returns something specifically for that

NoamDev:
No, wait that's still incorrect. Even if you're not the subject of the iteration you're still producing a recursive.

It just might be a boring recursive. At the very least this means we need to have a different Self for each variable I think?

This is a lot more complicated than I thought.

PR #77, opened 2025-08-05
check all selves are guarded

### LINKED ISSUE #62: Stack overflow SIGSEGV from bad recursive type
Par seems very cool! I started exploring it today and looking for interesting edge cases.

```
> cat bad.par
dec Loop : recursive self
def Loop = begin loop

> ./target/debug/par-lang run bad.par
fish: Job 1, './target/debug/par-lang run bad…' terminated by signal SIGSEGV (Address boundary error)
```

The stack trace contains many nested instances of `par_lang::par::types::Context::check_command::h166a013f71a081bc`.

### LINKED ISSUE #63: Poor diagnostic for bad iterative type
Similar but different from #62.

Image

```
dec Loop : iterative self
def Loop = begin loop
```

```
inet compilation error: UnguardedLoop(At { start: Point { offset: 43, row: 1, column: 17 }, end: Point { offset: 47, row: 1, column: 21 } }, None)
```

### LINKED ISSUE #68: Feature: Require guarding `self` of `recursive`/`iterative` by `either`/`choice`
**This issue describes an extension to the type system that should be implemented.**

The current type system places only one restriction on the usage of `self`: it needs to be _positive_ relative to the corresponding `recursive`/`iterative`.

This allows specifying `recursive` types like:

```
recursive (Nat) self
```

and `iterative` types like:

```
iterative (Nat) self
```

Neither are problematic from the soundness perspective, but:
- The `recursive` type isn't constructible, since it has no base variant
- The `iterative` type will cause a concurrent infinite loop, a kind of a leak

The latter will not cause a program getting stuck (except when it eventually runs out of memory), but is a pointless behavior regardless.

I have not found any valid use-case of `recursive`/`iterative` where the `self` isn't _guarded_ by an `either`/`choice` respectively.

**What does guarding mean?**

It means that:
- There must be at least one `either` between a `self` and its corresponding `recursive`
- There must be at least one `choice` between a `self` and its corresponding `iterative`

Some examples of valid types:

```
recursive either {  // this `either` guards the `self`
  .end!,
  .item(a) self,
}
```

```
iterative choice {  // this `choice` guards the `self`
  .close => !,
  .next => (a) self,
}
```

```
recursive choice {
  .close => !,
  .next => either {  // this `either` guards the `self`
    .end!,
    .item(a) self,
  }
}
```

However, this would not be a valid type:

```
recursive choice {
  .close => !,
  .next => (a) self,
}
```

Because there is no `either` between `recursive` and `self`.

**Solving this issue also solves #62 and #63**

NoamDev:
fixes #68 , fixes #62, fixes #63

### DIFF
--- a/src/par/types.rs
+++ b/src/par/types.rs
@@ -21,6 +21,8 @@ pub enum TypeError {
     DeclaredButNotDefined(Span, GlobalName),
     NoMatchingRecursiveOrIterative(Span),
     SelfUsedInNegativePosition(Span),
+    UnguardedRecursiveSelf(Span),
+    UnguardedIterativeSelf(Span),
     TypeNameNotDefined(Span, GlobalName),
     TypeVariableNotDefined(Span, LocalName),
     DependencyCycle(Span, Vec<GlobalName>),
@@ -289,7 +291,13 @@ impl TypeDefs {
             for param in params {
                 type_defs.vars.insert(param.clone());
             }
-            type_defs.validate_type(typ, &IndexSet::new(), &IndexSet::new())?;
+            type_defs.validate_type(
+                typ,
+                &IndexSet::new(),
+                &IndexSet::new(),
+                &IndexSet::new(),
+                &IndexSet::new(),
+            )?;
         }
 
         Ok(type_defs)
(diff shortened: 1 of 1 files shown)


faiface:
Very good! And good error messages. Merging too!

PR #78, opened 2025-08-05
Add check command to the CLI

NoamDev:
Usage: `par-lang check FILES`

Supports wildcards even in windows, so you can do: `par-lang check ./examples/*.par`


faiface:
Also great. Merging!

PR #81, opened 2025-08-07
Make Nat.Repeat lazy

NoamDev:


faiface:
Oh nice! But let's actually make it a separate function, at least for now. Like `Nat.RepeatLazy`.

PR #83, opened 2025-08-11
docs: Update documentation to reflect new language features

EatChangmyeong:
This PR updates a number of outdated documentation chapters to better reflect new language features introduced to Par.

## Summary of Changes

* Getting Started
  * Add `check` command
* Strings & Numbers
  * **Breaking change:** change chapter name to "Primitive Types". Existing links from outside will break.
  * Add link to Built-In Definitions chapter
  * Add `Byte` and `Bytes` types
  * Fix `Chars` function to handle the new `.end` payload
* Recursive / Iterative
  * Add descriptions about "guarded `self`" restriction

## Other Concerns

* Does the playground still display types for definitions? The Strings & Numbers chapter mentions that moving the cursor to definitions will show their types, but I've never seen this feature actually working.

### DIFF
--- a/docs/src/structure/primitive_types.md
+++ b/docs/src/structure/primitive_types.md
@@ -1,13 +1,15 @@
-# Strings & Numbers
+# Primitive Types
 
 Before taking a stroll in the diverse garden of Par's types, let's stop by the most basic ones:
 the _primitives._
 
-At the moment, Par has four primitive types:
+At the moment, Par has six primitive types:
 - **`Int`** — Integers, positive and negative whole numbers, arbitrary size.
 - **`Nat`** — Natural numbers, starting from zero, arbitrary size. They are a subtype of `Int`.
 - **`String`** — UTF-8 encoded sequence of Unicode characters.
-- **`Char`** — Singular Unicode characters.
+- **`Char`** — Singular Unicode character.
+- **`Byte`** — Singular data unit that consists of eight bits. They are a subtype of `Bytes`.
+- **`Bytes`** — Sequence of bytes.
 
 > There's a **significant distinction** between _primitives_ and all other types in Par.
 >
@@ -28,6 +30,9 @@ Primitives are manipulated using magical built-in functions.
    ```
 2. Press **Compile,** and **Run.** Scroll the list that pops up.
    ![List of built-in functions](../images/strings_and_numbers_1.png)
+
+Alternatively, the [Built-In Definitions](../builtin.md) chapter contains all built-in definitions
+available in the language.
 
 **To figure out the type of a built-in function:**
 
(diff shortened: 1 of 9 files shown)


docs/src/structure/primitive_types.md
```diff
@@ -1,13 +1,15 @@
...
 
-At the moment, Par has four primitive types:
+At the moment, Par has six primitive types:
 - **`Int`** — Integers, positive and negative whole numbers, arbitrary size.
 - **`Nat`** — Natural numbers, starting from zero, arbitrary size. They are a subtype of `Int`.
 - **`String`** — UTF-8 encoded sequence of Unicode characters.
-- **`Char`** — Singular Unicode characters.
+- **`Char`** — Singular Unicode character.
+- **`Byte`** — Singular data unit that consists of eight bits. They are a subtype of `Bytes`.
+- **`Bytes`** — Sequence of bytes.
```
faiface, line 12: I'd additionally stress that it's a contiguous (in memory) sequence of bytes. So:
> Contiguous (in memory) sequence of bytes.

docs/src/structure/primitive_types.md
```diff
@@ -182,7 +187,45 @@ copy-paste this one, if you ever need it:
...
+literals are written in decimal, enclosed in double angle brackets:
+
+```par
+def Byte1 = <<65>>   // inferred as `Byte`
+def Byte2 = <<321>>  // out-of-bounds values are automatically wrapped
+```
+
+Since `Byte` is a subtype of `Bytes`, every variable of type `Byte` can be used as a `Bytes`, too.
+
+Just like `Char`s, there's a built-in function to check if a `Byte` falls into a user-defined range:
```
faiface, line 207: I'd rephrase as:
> Just like for `Char`'s, there's a built-in function to check if a `Byte` is a part of a byte class. For `Byte`s, that's mainly byte ranges:

docs/src/structure/primitive_types.md
```diff
@@ -182,7 +187,45 @@ copy-paste this one, if you ever need it:
...
+
+```par
+def IsMsbSet = Byte.Is(<<192>>, .range(<<128>>, <<255>>)!)  // .true!
+```
+
+## `Bytes`
+
+`Bytes` are sequences of one or more bytes. Their literals are similar to those of `Byte`s, except
```
faiface, line 215: Add:
> They are laid out continguously in memory, in order to take the least possible amount of space.

docs/src/structure/primitive_types.md
```diff
@@ -182,7 +187,45 @@ copy-paste this one, if you ever need it:
...
+that multiple decimal values are allowed, and are delimited by spaces:
+
+```par
+def Bytes1 = <<65 91>>  // inferred as `Bytes`
+def Bytes2 = <<>>       // Error! Empty `Bytes` are not supported
```
faiface, line 220: Not true, empty `Bytes` are allowed, just apparently you need a space:
```
def Bytes2 = << >>
```
This works fine. Not sure at the moment why the parser fails without a space...

docs/src/structure/primitive_types.md
```diff
@@ -182,7 +187,45 @@ copy-paste this one, if you ever need it:
...
+```
+
+A `Bytes` can also be broken down to a list of `Byte`s:
+
+```par
+dec Bytes : [Bytes] List<Byte>
+def Bytes = [bs] Bytes.Reader(bs).begin.byte.case {
+	.end r => let ! = Result.Always(type !)(r) in .end!,
```
faiface, line 228: Actually, it's probably clearer to just drop the result here entirely:
```
  .end _ => ...
```

docs/src/structure/primitive_types.md
```diff
@@ -182,7 +187,45 @@ copy-paste this one, if you ever need it:
...
+	.byte(b) rest => .item(b) rest.loop,
```
faiface, line 229: Indentation in the docs is 2 spaces

docs/src/types/iterative.md
```diff
@@ -75,6 +75,20 @@ Just like [recursive types](./recursive.md), iterative types can be equated with
    ```
 4. And so on...
 
+> Just like [`recursive`](./recursive.md) types, `iterative` types have an important restriction:
+> **every `self` reference for an `iterative` must be guarded by a `choice`.** The `choice` doesn't
+> have to be right next to the `iterative`, but it *has* to be somewhere between `iterative` and
```
faiface, line 80: in-between

docs/src/types/recursive.md
```diff
@@ -136,6 +136,22 @@ The recursive type can be thought of as being equivalent to its _expansion_. Tha
 > This one starts with a [choice](./choice.md), which enables polling the elements on demand, or
 > cancelling the rest of the stream. However, being recursive, a `FiniteStream<a>` is guaranteed
 > to reach the `.end!` eventually, if not cancelled.
+>
+> There is nonetheless still an important restriction: **every `self` reference for a `recursive`
```
faiface, line 140: nonetheless ~~still~~

docs/src/types/recursive.md
```diff
@@ -136,6 +136,22 @@ The recursive type can be thought of as being equivalent to its _expansion_. Tha
...
+> must be guarded by an `either`.** The `either` doesn't have to be right next to the `recursive`,
+> but it *has* to be somewhere between `recursive` and `self`:
```
faiface, line 142: in-between

faiface (changes requested):
Honestly, amazing job! I suggested a small couple of stylistic changes.

> Does the playground still display types for definitions?

It does not currently but that's temporary. Actually I'm working right now on fixing it. Something broke with spans, I fixed one thing but it's still not entirely right. I'm prioritizing it, should have it fixed soon. So yeah, it is a feature, temporarily turned off :D

faiface:
@EatChangmyeong Actually I just fixed the showing of types in the playground!

faiface (approved):
Great! I just fixed the "in between" => "in-between", and I'm merging :)

PR #88, opened 2025-08-20
Store FileName info next to Span

coolreader18:
With this, type-hover actually works properly - before, it was not differentiating between the actual source file and `Builtin.par`. I considered putting `FileName` inside `Span` itself, but there was a lot of code that really expected `Span` to be `Copy` and this seemed easier.

I've been testing this with a local vscode extension that simply loads `par-lang lsp` as a language server, but I'm not exactly sure how it would fit into the repo if I were to want to commit it.


coolreader18:
I think the issue was that it's not just about changing the span of the names - e.g. hovering over column 21 of line 48 in `MiniGrep.Par`:
```
  .err e => do { Debug.Log(e) } in .end!,
//                  ^
```
shows Int, because line 48 in `Builtin.par` is
```
dec Int.Compare : [Int, Int] Ordering
//                  ^
```
But yeah, I'll remove that line from `qualify`.

faiface:
@coolreader18 Oh, there's a lot more of those erasures, in other places. Sorry I wasn't specific enough before. All in other `qualify()` methods.

Btw, I wonder if the behavior you're describing isn't due to not running an up-to-date version? Because I tackled this specific thing some days ago, precisely with that erasure that we're removing now. However, your solution is a lot better!

Btw, I texted you on Discord too.

So, could you please remove it everywhere else too? I think the PR is ready to merge after that.

PR #89, opened 2025-08-21
Make Char a subtype of String

KilianVounckx:
As discussed on
[discord](https://discord.com/channels/1336795503634939977/1336795504171941983/1408064248009068677),
this PR makes Char a subtype of String.

### DIFF
--- a/docs/src/builtin.md
+++ b/docs/src/builtin.md
@@ -131,7 +131,6 @@ dec Char.Is     : [Char, Char.Class] Bool
 /// String
 
 type String.Builder = iterative choice {
-  .addChar(Char) => self,
   .add(String) => self,
   .build => String,
 }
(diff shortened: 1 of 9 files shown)


faiface:
Looks great!

One detail: now that it's a subtype, I think it makes sense to do it like for Bytes/Byte, so a single character string literal would be inferred as Char.

That makes separate Char literals kinda redundant, so I'd suggest removing Char tokens, and instead treating both single and double ticks as string tokens. With the inferrence based on length.

Do you think you can add this too?

faiface:
Very good! Just one last thing: gotta update the "Primitive types" section of the docs.

Should be just a couple of small changes :) 

After that, ready to merge

KilianVounckx:
Commit  af79a69 already updates the primitive types docs. Or do you mean something else?

PR #92, opened 2025-08-24
[LSP] Get go-to-definition working for globals

coolreader18:
Getting it working for local variables is more complicated, but thanks to the nature of the relatively flat global namespace and the way local and global variables are syntactically distinct, go-to-def was fairly easy for globals.


faiface (approved):
LGTM! Thanks a lot for this addition! Merging

PR #97, opened 2025-09-06
fix windows using invalid type

Easyoakland:
Didn't look too deeply at this. I'm guessing that `ByteView` was valid at some point in the past but no longer exists. This seems to fix compilation failing because `Bytes` doesn't exist.


faiface:
Oh, yeah, didn't try on other platforms. Thanks a lot!

PR #99, opened 2025-09-07
Add button to automatically reload an opened file in the playground

Easyoakland:
Adds a checkbox which appears after loading a file in the playground.

Works by checking the mtime of the file every frame and reloading the contents if the file has been changed since last load.


faiface:
Sure, if this help you :D Mergin!

Issue #74, opened 2025-07-31
Feature: Reintroduce either & choice subtyping


faiface:
Before the current interaction network runtime, Par had subtyping of either and choice types.
It posed some problems for the initial implementation of the runtime, so we removed it and
started considering two either or two choice types subtypes only if they had the same branches,
and the payloads of those were subtypes.

Recently, I've modified the runtime to work with string based signal names, as opposed to indices,
and therefore there is no longer an obstacle to reintroducing the subtyping that was there
previously.

Here's how it's supposed to work:

## `either`

A either type `X` is a subtype of another either type `Y` iff every variant in `X` is
also present in `Y`, and its payload in `X` is a subtype of the corresponding
payload in `Y`.

For example:

```par
type X = either {
  .left Nat,
}

type Y = either {
  .left Int,
  .right String,
}
```

Here, every variant (only `.left`) of `X` is present in `Y` and its payload in `X` (`Nat`) is a subtype
of its corresponding payload in `Y` (`Int`).

## `choice`

A choice type `X` is a subtype of another choice type `Y` iff every method in `Y` is also present
in `X`, and its result in `X` is a subtype of the corresponding result in `Y`.

For example:

```par
type X = choice {
  .left => Nat,
  .right => String,
}

type Y = choice {
  .left => Int,
}
```

Here, every method (only `.left`) of `Y` is present in `X` and its result in `X` (`Nat`) is a subtype
of its corresponding result in `Y` (`Int`).

## Special cases: `either {}` and `choice {}`

An `either {}` is convertible to any type and any type is convertible to `choice {}`. Even with
the subtyping above, these conversions have to be performed manually.

While I initially rejected the idea, not having to do this conversion comes super handy when an
empty either or a choice is nested deep inside and a manual conversion involves re-constructing the
value.

Therefore, it makes sense to make:
- `either {}` a subtype of any type
- any type a subtype of `choice {}`

PR #100, opened 2025-09-14
feat: special cases for either/choice subtyping

tannerduve:
Added short guards to ensure that `either {} <: other` and `self <: choice {}` for all other, self. added a test case for each.


faiface:
Very good! Thank you very much. One step closer to resolving the either/choice subtyping #74

Merging

PR #102, opened 2025-09-28
Rewrite subtyping

NoamDev:
Subtyping is now based on [A Logical Account of Subtyping for Session Types](https://pureportal.strath.ac.uk/files/182097742/Horne-Padovani-EPTCS-2023-A-logical-account-of-subtyping-for-session-types.pdf)

Currently the performance isn't the best, but it's good enough in release mode. If it ever becomes more of a priority, we can optimize.

### DIFF
--- a/src/par/types/assignability.rs
+++ b/src/par/types/assignability.rs
@@ -1,33 +1,113 @@
 use crate::location::Span;
-use crate::par::language::LocalName;
+use crate::par::types::assignability::SubtypeResult::{Compatible, Cycle, InCompatible};
 use crate::par::types::{PrimitiveType, Type, TypeDefs, TypeError};
-use std::collections::{BTreeMap, HashMap, HashSet};
-use std::rc::Rc;
-use std::{env, fmt::Write as _};
+use indexmap::IndexSet;
+use std::cmp::max;
+use std::collections::BTreeMap;
+use std::env;
+use std::ops::BitAnd;
 
-#[derive(Debug)]
-enum FixPointType {
-    Recursive,
-    Iterative,
+#[derive(Clone)]
+struct SubtypeContext<'a> {
+    type_defs: &'a TypeDefs,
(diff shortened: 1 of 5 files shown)


src/par/types/assignability.rs
```diff
@@ -1,33 +1,119 @@
...
+                self.normalize(self.type_defs.get(&span, &name, &args)?.dual(Span::None))?
+            }
+            t => t,
+        })
+    }
+}
+
+enum SubtypeResult {
+    Compatible,
+    InCompatible,
```
faiface, line 38: Incompatible is a single word, no need for the capitalization on C

src/par/types/assignability.rs
```diff
@@ -1,33 +1,119 @@
...
+    Cycle {
+        min_left: Type,
+        size_left: u32,
+        min_right: Type,
+        size_right: u32,
+        ttl: usize,
```
faiface, line 44: This TTL idea needs explanation here, so that other people can extend this algorithm if necessary for new features without breaking this part.

src/par/types/expansion.rs
```diff
@@ -113,4 +123,36 @@ impl Type {
...
+                asc,
+                label,
+                body,
+            } => Self::expand_iterative(span, asc, label, body),
+            _ => Ok(self.clone()),
+        }
+    }
+
+    // This variant does not make sure iterative's asc isn't empty
+    pub fn expand_fixpoint_unsafe(&self) -> Result<Self, TypeError> {
```
faiface, line 144: `unsafe` already has a meaning in Rust, let's call this something else, like... `ignoring_asc`

faiface (changes requested):
Incredible job, and I'm also very happy with the LOC reduction! The algorithm is overall very well and clearly written, just a couple of things to change.

Especially, we need an explanation on the TTL idea.

faiface (approved):
All good now! Merging

PR #103, opened 2025-09-30
either/choice type inference

NoamDev:


src/par/types/lattice.rs
```diff
@@ -0,0 +1,231 @@
...
+            span.clone(),
+            Box::new(union_types(typedefs, span, inner1, inner2)?),
+        ),
+        (Type::DualBox(_, inner1), Type::DualBox(_, inner2)) => Type::DualBox(
+            span.clone(),
+            Box::new(intersect_types(typedefs, span, inner1, inner2)?),
+        ),
+        (Type::Break(_), Type::Break(_)) => Type::Break(span.clone()),
+        (Type::Continue(_), Type::Continue(_)) => Type::Continue(span.clone()),
+        (Type::Primitive(_, p1), Type::Primitive(_, p2)) if p1 == p2 => {
```
faiface, line 51: Handling of primitives here doesn't take into account their subtyping. For example `union{Nat,Int}` should be `Int`, and `intersect{Nat,Int}` should be `Nat`. Perhaps some `min`/`max` (or `union`/`intersect`) functions on the primitive types would be nice, also usable in other places where this is of concern (multiple places).

faiface (approved):
All good now!

PR #109, opened 2025-10-16
Fix byte literal parsing

EatChangmyeong:
Fixes a parser bug that rejects the empty byte literal `<<>>` (turns out that this sequence was tokenized into `Lt`-`Link`-`Gt` instead of `Lt`-`Lt`-`Gt`-`Gt`).

### DIFF
--- a/docs/src/structure/primitive_types.md
+++ b/docs/src/structure/primitive_types.md
@@ -219,8 +219,7 @@ that multiple decimal values are allowed, and are delimited by spaces:
 
 ```par
 def Bytes1 = <<65 91>>  // inferred as `Bytes`
-def Bytes2 = << >>      // zero-byte sequence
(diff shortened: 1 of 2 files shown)


EatChangmyeong:
> Nice! I wonder if it's possible to fix the lexer bug instead of a workaround in the parser

I can come up with two ways:

* Remove `TokenType::Link` and tokenize everything into simple `Lt` and `Gt`. This looks cleaner, but it'll allow awkward broken link operator (`< >`). Maybe this is not a problem because broken byte literals are already allowed?
* Add `<<` and `>>` as new tokens. Unfortunately this will certainly trigger the [angle bracket problem](https://en.cppreference.com/w/cpp/language/angle_bracket_hack.html)

NoamDev:
Alright, yeah, this is more difficult than I thought

PR #113, opened 2025-11-17
Fixed docs links in Commands

NeonOxide:
There were some relative links in the Commands section of the docs that were broken I guess due to moving the files without updating the relative locations. 

I fixed them by substituting the broken links with links containing the location from the root (in order to avoid them breaking again in case of just moving the files), but if you want to keep it relative, the change is quite simple to make.

I have not checked the rest of the files to see if there are other broken links as I haven't had the time. Thus, I cannot ensure that the rest of the docs do not contain other broken links. Thanks in advance for your time :D.

### DIFF
--- a/docs/src/processes/commands/looping_and_branching.md
+++ b/docs/src/processes/commands/looping_and_branching.md
@@ -35,7 +35,7 @@ see how it works out!
 
 ```par
 def Concat = [strings] do {
-  // code continued below.
+  // code continued below...
 ```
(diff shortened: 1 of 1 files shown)


NeonOxide:
Seems like gh-pages does not like absolute paths. I will just do the correct relative links and reopen a PR. Sorry for the oversight. Although, as an old CS saying goes, it worked on my machine!

PR #137, opened 2025-11-30
add more builtins

Defelo:
This patch adds builtin definitions to read command line arguments and environment variables.

For `Os.Env` I first tried to make it a `BoxMap<Bytes, Bytes>` using the `provide_boxmap` function, but I couldn't figure out how to construct the value `Handle`s in the `map` parameter, so that's why I chose to simply use a `List<(Bytes) Bytes>` for now. Also, in Rust [`std::env::vars_os()`](https://doc.rust-lang.org/stable/std/env/fn.vars_os.html) gives you an iterator over key value pairs anyway, so I think a `List` should be fine as well.

Ok, actually I think I just did find a way to construct these `Handle`s now (which at least *seems* to work), but I have no idea if I'm missing something here. If I add the following `make_pair` function to `Handle`, I can use it to create a `(Handle, Handle)` pair for every map item, call `provide_bytes()` on one `Handle` and put the other one into the map.

```rust
    pub fn make_pair(&self) -> (Handle, Handle) {
        let (a, b) = self.net.lock().expect("lock failed").create_wire();
        let a = Handle {
            net: self.net.clone(),
            tree: Some(a),
        };
        let b = Handle {
            net: self.net.clone(),
            tree: Some(b),
        };
        (a, b)
    }
```

Now, since honestly I don't really know what I'm doing here, I will stick with the `List` implementation for now, but I would really appreciate it if you could have a look and maybe tell me if there's anything wrong with this approach or if I'm missing something and whether the `BoxMap` (or `Map`?) approach would make more sense than a simple `List` at all :slightly_smiling_face:

### DIFF
--- a/src/par/builtin/Builtin.par
+++ b/src/par/builtin/Builtin.par
@@ -308,6 +308,7 @@ dec Os.TraverseDir : [Os.Path] Result<Os.Error, recursive either {
   .dir(Os.Path, self) self,
 }>
 
+dec Os.Env : List<(Bytes) Bytes>
 
 /// Url
 
--- a/src/par/builtin/os.rs
+++ b/src/par/builtin/os.rs
@@ -156,6 +156,15 @@ pub fn external_module() -> Module<std::sync::Arc<process::Expression<()>>> {
                 ),
                 |handle| Box::pin(os_traverse_dir(handle)),
             ),
+            Definition::external(
+                "Env",
+                Type::name(
+                    Some("List"),
+                    "List",
+                    vec![Type::pair(Type::bytes(), Type::bytes())],
+                ),
+                |handle| Box::pin(os_env(handle)),
+            ),
         ],
     }
 }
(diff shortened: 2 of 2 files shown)


faiface (changes requested):
I'm very sorry for keeping you waiting so long!

The list approach is good, it's not necessary for the `Os.Env` to return a map.

However, one thing I'd change is to return strings instead of bytes! Why return bytes in the first place? :D Even Rust's `std::env::args` and `std::env::vars` provide strings.

So please change the types to
```
dec Os.Args : List<String>
dec Os.Env  : List<(String) String>
```

And notice, that after this change, getting a boxmap will even be trivial!
```
let env = BoxMap.String(type String)(Os.Env)
```

That's all, let's change these types (including the Rust implementation), and we can merge then.

faiface:
Oh, I see! Yeah, then making a BoxMap/Map with a Bytes key type would be the best solution here, and otherwise keeping your implementation.

Good point there about the args being useless now :D Adding such a context is not a trivial task, so feel free to choose one of these options:
- Remove `Os.Args` for now, and we'll come back to it later or in another PR. Plus add bytes keyed maps.
- Try and add the `Context`, or some other way to be able to supply dynamic information to the runtime

If you feel like the second option, feel free! But I'm very good with the first option as well.

Issue #126, opened 2025-11-23
[Feature] Implicit generic arguments


faiface:
Generics in Par are fairly powerful. There are three places that involve type variables:
- Type aliases themselves: `type List<a> = ...`
- Forall types: `dec List.Map : [type a, b] ...`
- Exists types: `type SetModule<a> = (type set) ...`

The case of type aliases is different than the latter two. A generic type alias is not a first-class citizen, a `List` is nothing on its own. Instead, it's a meta-programming feature, allowing you to share a common template for many types. Unless we involve higher kinds, they are fine as they are.

The latter two, forall and exists types are something else. They are types of real values, and they are first-class. Unlike in many languages, it's perfectly fine to store a generic value in a variable, or pass it as an argument to a function:

```par
let id = box [type a] [value: a] value

let str = id(type String)("Hey")  // = "Hey"
let num = id(type Int)(42)        // = 42
```

From a theory perspective, Par's forall and exists types are the famous System F with existential types. Which is really neat and powerful.

**However,** using them in practice often proves unwieldy and cluttered. Let's take the usual list map function as an example:

```par
dec List.Map : [type a, b] [List<a>, box [a] b] List<b>
```

Using it becomes very wordy:

```par
let numbers: List<Int> = *(1, 2, 3, 4, 5, 6, 7, 8, 9)
let strings = List.Map(type Int, String)(numbers, box Int.ToString)
```

And that's a mild example! Let's take the `GrepFile` function from `examples/MiniGrep.par`:

```par
dec GrepFile : [String, Os.Reader] Iterator<Os.Error, (Nat) String>
def GrepFile = [needle, reader] do {
  let parser = String.ParserFromReader(type Os.Error)(reader)
  let lines = Lines(type Os.Error)(parser)
  let lines = Enumerate1(type Os.Error, String)(lines)
} in Filter(type Os.Error, (Nat) String)(lines, box [(_) line] Contains(needle, line))
```

Now let's take a look what it **could** look like without all the boilerplate:

```par
dec GrepFile : [String, Os.Reader] Iterator<Os.Error, (Nat) String>
def GrepFile = [needle, reader] do {
  let parser = String.ParserFromReader(reader)
  let lines = Lines(parser)
  let lines = Enumerate1(lines)
} in Filter(lines, box [(_) line] Contains(needle, line))
```

So much more readable!

# What's so hard? Hindley-Milner solved it decades ago!

Let's entertain the usual approach for a moment:
[...]

PR #140, opened 2025-12-02
builtin: change `Os.Env` to return a readonly `BoxMap`

DeedleFake:
#137's `Os.Env` immediately made me want an easy way to find a specific variable. I'm not sure if the `List.Find` implementation is the correct way to do that in the general case, as it seemed to require a fair bit of `box` usage which limits it a bit.

Speaking of which, should there be a `Bytes.Equals`? I can add that in, too, if you want.

### DIFF
--- a/src/par/builtin/Builtin.par
+++ b/src/par/builtin/Builtin.par
@@ -386,11 +386,11 @@ type BoxMap<k, v> = iterative box choice {
   .delete(k) => self,
 }
 
-type BoxMap.Readonly<k, v> = iterative box choice {
-	.size => Nat,
-	.keys => List<k>,
-	.list => List<(k) box v>,
-	.get(k) => Option<box v>,
+type BoxMap.Readonly<k, v> = box choice {
+  .size => Nat,
+  .keys => List<k>,
+  .list => List<(k) box v>,
+  .get(k) => Option<box v>,
 }
 
 dec BoxMap.String : [type v] [List<(String) box v>] BoxMap<String, v>


DeedleFake:
I'm looking forward to #126, as calling generic functions right now is not only verbose, it also clashes awkwardly with `->`. You wind up with things like `Os.Env->{List.Find(type (Bytes) Bytes)}(box ...)`.

faiface:
@DeedleFake Hey, also sorry here for a late reaction, been busy :D 

This is some good effort, and there certainly is an appetite for more standard library functions!

However, like you pointed out yourself, generic functions that are not like constructors (like say `List.Builder`), are currently awkward to use because they don't play well with `->` and also they're unnecessarily verbose.

Sometimes specifying the explicit argument is useful and normal, like with the constructors, but on things like `List.Map` or your `List.Find`, it's kinda embarrassing to get it into the standard library :D 

Fortunately, @NoamDev is working on fixing that with the implicit generics feature! So we should actually have that fairly soon!

Another thing about something like `List.Find` is that it's clearly a part of a larger picture of list functions. It's kinda lonely being there alone. So once we get the implicit generics, I think we'll just need to add a whole bunch of list functions!

So I'd wait until then.

For the rest, the `Bytes.Equals` and `Bytes.Compare` functions are super useful! Let's keep those.

The `Os.EnvVar`, I'm kinda leaning towards building a `BoxMap`. You do it once and then you can get any environment variable quickly. I'd say I'd also skip `Os.EnvVar` but later turn `Os.Env` into a `BoxMap` as such, so then one can just do `Os.Env.get` or `Os.Env.list` as they please.

What do you think? If you agree, let's just keep the `Bytes` functions and I'll merge :)

faiface:
@DeedleFake Right, the `BoxMap` type is essentially just an interface, it doesn't prescribe that the instance has to hold a map, so `.get` can literally just do what your current `Os.EnvVar` does.

But I'm a little unsure about setting... is setting env vars even concurrency safe? I'd probably skip that one atm and added a type like this:
```
type ReadonlyMap<k, v> = box choice {
  .size => Nat,
  .keys => List<k>,
  .list => List<(k) box v>,
  .get(k) => Option<box v>,
}
```
which would naturally be a super-type of `BoxMap`.

And have the `Os.Env` def provide this. What do you think?

faiface:
Very nice, I like the name `BoxMap.Readonly`.

Just one thing, the `iterative` is actually unnecessary there since `self` is not referenced.

Once you remove that, I'm merging :)

faiface:
I'm gonna merge when I'm at my laptop.

DeedleFake:
> At least on Linux, I don't think it's actually possible to change the environment of a running process.

It is. That's what the previous conversation was about: https://github.com/faiface/par-lang/pull/140#issuecomment-3608511214 Not only is it possible, but because of the way libc reads environment variables it's extremely unsafe to do if there's pretty much any kind of concurrency. As long as anything anywhere in the process uses libc to read an environment variable, changing a variable sometime between that read happening and the returned pointer being used, even if that write is done without using libc, will cause major problems.

DeedleFake:
> Sorry, I meant "change the environment of a running process from the outside". So, what I wanted to say was, if we never change the environment ourselves, we can expect it to always be the same, so we only have to load it once.

Yeah, I don't think that's possible, at least not under normal circumstances. But imagine if someone uses Rust to wrap a C library that, buried somewhere in it, is a `setenv()` call. If the environment is only read once, that would cause it to be inaccurate, and, even weirder, it could cause differences between the return values of `.list` and `.get`, since `.get` just bypasses the list and does a direct lookup every time. Rust's implementation is not prone to the same problems, mostly, that reading an environment variable using libc is since it makes copies, but a cache could still get out of date.

PR #144, opened 2025-12-03
builtin: add more functions

DeedleFake:
Adds more builtin functions and types, such as common list operations, string operations, etc., as mentioned in https://github.com/faiface/par-lang/pull/140#issuecomment-3608340331. This is not intended to be merged until after #126 is done.

### DIFF
--- a/src/par/builtin/Builtin.par
+++ b/src/par/builtin/Builtin.par
@@ -266,7 +266,10 @@ def List.All = <a>[list] [f]
   }
 
 dec List.Any : <a>[List<box a>] [box [a] Bool] Bool
-def List.Any = <a>[list] [f] List.All(list, box [v] Bool.Not(f(v)))
+def List.Any = <a>[list] [f] List.All(list, box [v] f(v).case {
+  .true! => .false!,
+  .false! => .true!,
+})
 
 dec List.Dedup : <a>[List<box a>] [box [a, a] Bool] List<box a>
 def List.Dedup = <a>[list] [eq]


DeedleFake:
I tried to implement `List.Min` and `List.Max` using `List.MinMax`, but I couldn't figure out how to get it not to complain about not using the other value that's returned. I _could_ just copy and paste `List.MinMax` and remove the other one, I guess.

NoamDev:
@DeedleFake If you accept `box a`, then you should return `box a` to be more permissive.

That's because box a can be implicitly converted to a, but not the other way around.

box a is basically a but with the added information that it's replicable and droppable.

If for example it was that way in MinMax, then you'd be able to drop the min or max.

DeedleFake:
That makes sense. By not returning it, it's changing type and losing that extra functionality. I'll change it.

DeedleFake:
@NoamDev

The docs very specifically say that it should _not_ return a `box a`: https://faiface.github.io/par-lang/types/box.html Is that a mistake in the docs?

faiface:
@DeedleFake Regarding the `List.Unzip`, you've actually stumbled upon a crucial limitation in the Par's structure! Unfortunately it's nor currently possible to do it that way, and won't be until we introduce some new language features, that I'm not sure what they should look like yet.

But essentially you need something like this:
```
def List.Unzip = [type a] [list] chan consumer {
  consumer(let yield1)
  let yield2 = consumer
  // produce items alternatingly
}
```
The `(let yield1)` is a made-up syntax here.

To see why this is problematic, imagine you're doing the same, but not returning just lists, but instead returning a pair of these iterators:
```
type Iter<a> = recursive choice {
  .next => either {
    .end!,
    .item(a) self,
  }
}
```
And then you'd do:
```
// we have yield1 and yield2 like above
yield1.case {
  .next => {
    yield1.item(...)
  }
}
// alternatingly produce on yield1 and yield2 in the same fashion
```
Well, if you look closely, now you've got a possible deadlock on the other side. Since the unzip returns two supposedly independent iterators, the consumer can call `.next` on the second iterator first, but the producer is waiting on the first yielder.

So having what's needed for the `List.Unzip`, we've only needed to change the list to an iterator to get a deadlock.

The key difference between the two here is that the iterator has a "feedback". For a list, the construction would not lead to deadlocks, but it's unclear how to exactly distinguish these two in the type system and enable this kind of a pattern without opening up deadlocks.

But, this and other uses of similar patterns are useful enough that something to tackle this will likely make it into the language eventually. For now, the `List.Unzip` will have to use builders.

faiface:
@DeedleFake 

> I'm not 100% sure that List.Result is the right to handle a list of things

Oh that's a very neat pattern there with the `else try!` :D 

The `List.Result` type is actually a useful one!

However, I'd say the better approach would a more boring one:
```
type List.Result<e, a> = recursive either {
  .end Result<e, !>,
  .item(a) self,
}
```
This still allows a similar pattern, just without else, well:
```
listResult.begin.case {
  .item(x) => { ... }
  .end try! => { ... }
}
```

faiface:
@DeedleFake 

> but the second one that you have there isn't valid because the -> operator will affect the first ()

Ah, I see what you mean! It works well if you write it like this:
```
dec List.Map : <a>[List<a>] [type b] [box [a] b] List<b>
```
This is a little awkward without using the pipe syntax
```
List.Map(numbers)(type String)(box Int.ToString)
```
but works great with the pipe syntax
```
numbers->List.Map(type String)(box Int.ToString)
```

faiface:
@DeedleFake I'm ready for merging the incubator branch, so I'm going to merge this PR and apply the remaining suggested changes myself :) Thank you so much for this contribution, much appreciated!

PR #152, opened 2025-12-08
runtime: implement a cache-friendly "V3" runtime.

FranchuFranchu:
This PR implements a more cache-friendly runtime, as a part of #153. This is the third runtime Par has, that's why it's called V3.

The `run`, `check` and `test` subcommands got a `-frt-v3` flag to run under the new runtime. `-frt-v2` runs the old runtime. The old runtime is used by default.

This PR should not break the v2 runtime. If a program that previously worked doesn't work, that's a bug and should be reported. I'd like this PR to be merged after all of those bugs are fixed; we shouldn't merge regressions into `main`. So please, test this branch thoroughly. 

In the future, we might remove the V2 runtime.

### Introduction

The runtime is still based on *interaction nets*. [Interaction nets](https://en.wikipedia.org/wiki/Interaction_nets) (inets) are a model of computation where programs are represented by graphs (a *net*). Each node corresponds to an operation or a value.

Nodes are connected to other nodes with edges through ports, and they have a *principal port* which determines what value it interacts with.

Par is particularly well-suited for interaction nets because it makes no distinction between commands and values. In the runtime, values and commands are represented by the same kind of thing.

Other programming languages that have inet runtimes are [Vine](https://vine.dev/) and [HVM](https://github.com/HigherOrderCO/HVM3/)

The runtime that preceded this one was also based on interaction nets. It used an approach where the many different types of commands and values in Par were compiled to a small set of nodes. This is the approach also used by other inet languages. This is very simple to implement and work with but it suffers from poor cache locality. Additionally, serialization and debugging is very hard.

However, it turns out that in Par we can do much better. Linear logic heavily restricts where and how duplication can take place. The older approach was "too general"; we can use Par's restrictions to implement a new runtime.

The new runtime stores the static program in an arena defined in `arena` module. The arena's
contents don't change while it's executed. The program state, which changes while it's running
is stored in different locations depending on how it's used. Some state is reference-counted while
state that is guaranteed to be used linearly is stored in a shared buffer that is allocated when
each definition is instantiated.

### Codebase changes

Many changes to the rest of the codebase had to be made for this runtime..

Currently, the new runtime is transpiled from the old runtime since the representations are very similar, to save some effort into refactoring the compiler.

A few changes had to be made to the old runtime and compiler to allow for this. Now, the old runtime distinguishes between `Par` and `Times`, and `Continue` and `Break`. This means that polarity errors on builtins now error, when previously they were not detected since `Continue` and `Break` were represented by the same node. **Porting old builtins might result in panics which are due to polarity type errors**

In the long term, this should not be necessary and be replaced direct compilation to the new runtime.

The old `Handle` was replaced by a `Handle` that can be either backed by the old runtime or the new runtime. Positive commands are `async` now; `send()`, `break_()` and `signal()`. **Old builtins will need extra `.await`s to work with the new runtime**

## TODOs

These should be done before merging

- [x] Test thoroughly
- [x] Remove debugging logs

These can be done in a new PR.

- [ ] Add support for boxes
- [ ] Add statistics
- [ ] Add support for nonlinearity with IO; `Handle::duplicate()` and `Handle::erase()` are `todo!()`s right now
- [ ] Add support for the playground. This would require making TypedHandle independent of the backing Handle.
- [ ] Make the new runtime the primary runtime :)

### Documentation

There are a doc comments in the new runtime code.


faiface:
**BUG**

When I run the `WebServer.par` example and load the website, I get a panic:
```
faiface@Michals-MacBook-Air-2 par-lang % par-lang run -frt-v3 examples/WebServer.par
Directory to serve: .
Starting server...

thread 'tokio-runtime-worker' (1227539) panicked at src/runtime/new/runtime.rs:444:47:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

thread 'tokio-runtime-worker' (1227539) panicked at src/runtime/new/readback.rs:39:61:
Sender dropped!: RecvError(())
```

faiface (approved):
The code is overall really good here! Amazing job @FranchuFranchu . There are some warnings remaining from Rust, but I'm gonna merge into the `incubator` branch so we can be testing all new things together, and we'll fix the remaining bugs and warnings there.

PR #154, opened 2025-12-08
Add implicit generics

### LINKED ISSUE #126: [Feature] Implicit generic arguments

NoamDev:
closes #126 description is there.
Note: The performance currently is not ideal on debug-mode.
Note: the generic argument would be inferred to be a box where relevant

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@
 
 ## 🚀 Get started with the [documentation](https://faiface.github.io/par-lang/introduction.html)
 
+## 🛠️ [VSCode extension](https://marketplace.visualstudio.com/items?itemName=par-lang.par-vscode)
+
 ## 📺 [What’s a linear programming language like? — Coding a “Mini Grep” in Par](https://youtu.be/nU7Lt6k3lNQ?si=8KtbTQMgp6VhZ5UP)
 
 ## 💡 Check out the [examples](examples/)
--- a/src/icombs/compiler.rs
+++ b/src/icombs/compiler.rs
@@ -190,9 +190,8 @@ pub struct Compiler {
     net: Net,
     context: Context,
     type_defs: TypeDefs,
-    definitions: IndexMap<GlobalName, Definition<Arc<Expression<Type>>>>,
+    definitions: IndexMap<GlobalName, (Definition<Arc<Expression<Type>>>, Type)>,
     global_name_to_id: IndexMap<GlobalName, usize>,
-    id_to_ty: Vec<Type>,
     id_to_package: Vec<Net>,
     lazy_redexes: Vec<(Tree, Tree)>,
     compile_global_stack: IndexSet<GlobalName>,
@@ -221,10 +220,9 @@ pub(crate) fn multiplex_trees(mut trees: Vec<Tree>) -> Tree {
 impl Compiler {
     fn compile_global(&mut self, name: &GlobalName) -> Result<TypedTree> {
         if let Some(id) = self.global_name_to_id.get(name) {
-            let ty = self.id_to_ty.get(*id).unwrap().clone();
             return Ok(TypedTree {
                 tree: Tree::Package(*id),
-                ty,
+                ty: Type::Break(Span::None),
             });
         };
         if !self.compile_global_stack.insert(name.clone()) {
@@ -234,7 +232,7 @@ impl Compiler {
             });
         }
         let global = match self.definitions.get(name).cloned() {
-            Some(def) => def.expression,
+            Some((def, _typ)) => def.expression,
             _ => return Err(Error::GlobalNotFound(name.clone())),
         };
 
@@ -306,7 +304,6 @@ impl Compiler {
         let old_net = core::mem::take(&mut self.net);
         let old_lazy_redexes = core::mem::take(&mut self.lazy_redexes);
         // Allocate package
-        self.id_to_ty.push(Type::Break(Span::default()));
         self.id_to_package.push(Default::default());
         let mut tree = self.with_captures(&Captures::default(), |this| f(this, id))?;
 
@@ -350,7 +347,6 @@ impl Compiler {
             .redexes
             .append(&mut core::mem::take(&mut self.lazy_redexes).into());
         self.net.assert_valid();
-        *self.id_to_ty.get_mut(id).unwrap() = tree.ty.clone();
         *self.id_to_package.get_mut(id).unwrap() = core::mem::take(&mut self.net);
         self.lazy_redexes = old_lazy_redexes;
         self.net = old_net;
@@ -439,16 +435,16 @@ impl Compiler {
         self.use_var(&Var::Name(name.clone()), usage, in_command)
     }
 
-    fn create_typed_wire(&mut self, t: Type) -> (TypedTree, TypedTree) {
+    fn create_typed_wire(&mut self) -> (TypedTree, TypedTree) {
         let (v0, v1) = self.net.create_wire();
         (
             TypedTree {
                 tree: v0,
-                ty: t.clone(),
+                ty: Type::Break(Span::None),
             },
             TypedTree {
                 tree: v1,
-                ty: t.dual(Span::None),
+                ty: Type::Break(Span::None),
             },
         )
     }
@@ -468,31 +464,6 @@ impl Compiler {
         else_branch: Option<usize>,
     ) -> Tree {
         Tree::Choice(Box::new(ctx_out), Arc::new(branches), else_branch)
-    }
-
-    fn normalize_type(&mut self, ty: Type) -> Type {
-        match ty {
-            Type::Name(loc, name, args) => {
-                let ty = self.type_defs.get(&loc, &name, &args).unwrap();
-                self.normalize_type(ty)
-            }
-            Type::DualName(loc, name, args) => {
-                let ty = self.type_defs.get_dual(&loc, &name, &args).unwrap();
-                self.normalize_type(ty)
-            }
-            Type::Box(_, inner) => self.normalize_type(*inner),
-            Type::DualBox(_, inner) if inner.is_positive(&self.type_defs).unwrap() => {
-                self.normalize_type(inner.clone().dual(Span::None))
-            }
-            Type::Recursive {
-                asc, label, body, ..
-            } => self.normalize_type(Type::expand_recursive(&asc, &label, &body).unwrap()),
-            Type::Iterative {
-                asc, label, body, ..
-            } => self
-                .normalize_type(Type::expand_iterative(&Span::None, &asc, &label, &body).unwrap()),
-            ty => ty,
-        }
     }
 
     fn compile_expression(&mut self, expr: &Expression<Type>) -> Result<TypedTree> {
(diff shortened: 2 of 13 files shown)


faiface:
@DeedleFake Yep, this is a crucial limitation that's also a part of the design. Essentially, when we infer types, we want to infer them in-place. We don't want unsolved variables keep hanging around. And so, the inference only considers the first argument to the function. That's what the syntax also indicates: you write `<a>[a] [a] ...` and not `<a>[a, a] ...`. As soon as we'd start considering more than the first argument, you start needing to keep unsolved type variables around in types.

So yeah, this isn't really getting fixed, that's how it's supposed to work.

What could help for your specific example would perhaps be some kind of syntax for specifying that a number literal should be considered an `Int` even though it can be a `Nat`.

faiface:
@DeedleFake Btw, for a more extreme example of this, try this 😆 
```
List.Concat(*(), *(1, 2, 3))
```

PR #174, opened 2026-02-19
update rust edition to 2024

NoamDev:
Updates the rust edition to 2024.
This enables [if-let chains and such](https://rust-lang.github.io/rfcs/2497-if-let-chains.html)

All the code changes were cerated by `cargo fix --edition`, so it might be safe to skip reviewing, given that it still builds and tests pass.
The changes were not strictly necessary, but they're a nice touch so why not.
The changes to imports are `cargo fmt` who the heck know why :D


faiface (approved):
Nice, thank you! Merging

Issue #156, opened 2025-12-10
command syntax pipe inconsistency with variable used as subject and argument


DeedleFake:
The docs say that `n->F` in command syntax is the same as `let n = F(n)`, but attempting to do that with the variable as both the subject and an argument fails despite it working when using `let` manually.

```par
def Test: ! = do {
	let n = 2

	let n = n->Nat.Mul(n) // Works.
	n->Nat.Mul(n) // Error: Variable `n` does not exist.
} in !
```

faiface:
@NoamDev The optimization would be more for the compiled code readability, but it might not be a problem

PR #175, opened 2026-02-21
Allow subject to appear in a command's payload

NoamDev:
fixes issue #156


faiface (changes requested):
It's good! Just some very minor change requests + I noticed we get a warning for an unused `use` (not introduced here... slipped), maybe we should remove it here.

Very good job @NoamDev, thank you so much!

faiface (approved):
All good now! Merging

PR #176, opened 2026-02-26
fix race condition

NoamDev:


faiface (approved):
Looks very good, thank you so much! Merging

PR #178, opened 2026-02-28
Refactor runtime into its own crate

NoamDev:
Note that this is a series of commits that are big refactors, after each one I made sure `cargo check` passes.
So you might want to review each commit separately, but that's up to you.


faiface (approved):
Looks really good and a step in the right direction! Thank you so much @NoamDev, ready to merge it.

PR #179, opened 2026-03-02
add web support

NoamDev:
To test it do:
```
Bash
rustup target add wasm32-unknown-unknown
cargo install --locked trunk
trunk serve
```

Works on firefox/zen. Doesn't work on chrome/edge. Gives `Maximum call stack size exceeded`. And it does not seem to matter how big I make the stack.


faiface (approved):
Absolute madman, looks great! Merging

PR #182, opened 2026-03-22
Separate compilation

NoamDev:


faiface (changes requested):
The code looks good on the first look. Just got two comments to functionality:
```
par-lang run-vm Fibonacci
```
Does not work, have to use `Fibonacci.Main`. But it does work with `par-lang run Fibonacci`, if you don't specify a def you get Main. I guess these two should align.

The second is, the generated `.pvm` file contains a lot of stuff like this:
```
�S�Loop body at At { start: Point { offset: 668, row: 30, column: 23 }, end: Point { offset: 923, row: 39, column: 5 }, file: FileName("./src/Queue.par") }|�T�T�Loop body at At { start: Point { offset: 627, row: 29, column: 15 }, end: Point { offset: 1082, row: 46, column: 3 }, file: FileName("./src/Queue.par") }|�U�U�Loop body at At { start: Point { offset: 532, row: 24, column: 5 }, end: Point { offset: 1084, row: 47, column: 1 }, file: FileName("./src/Queue.par") }|�V�V�Loop body at At { start: Point { offset: 532, row: 24, column: 5 }, end: Point { offset: 1084, row: 47, column: 1 }, file: FileName("./src/Queue.par") }|{W|W�Loop body at At { start: Point { offset: 627, row: 29, column: 15 }, end: Point { offset: 1082, row: 46, column: 3 }, file: FileName("./src/Queue.par") }|�W�W�Loop body at At { start: Point { offset: 668, row: 30, column: 23 }, end: Point { offset: 923, row: 39, column: 5 }, file: FileName("./src/Queue.par") }|�W�WQueue.PlayWithQueue| X!X RockPaperScissors.CalculateRound|�Z�ZockPaperScissors.StartGame~�Y�Y�Loop body at At { start: Point { offset: 733, row: 45, column: 24 }, end: Point { offset: 1228, row: 70, column: 3 }
```
I think these should not be there? It's not like they are parseable.

PR #183, opened 2026-04-02
Fix Request interact priority

### LINKED ISSUE #180: BUG: panic in the flat runtime
Running the following using the `run` command causes a panic.
```
type Church = box <a>[a] [box [a] a] a

dec Zero: Church
def Zero = box <a>[x] [f] x

def Main:! =
  do {
  let n = Zero
  let n = Bench.BlackBox(type Church)(n)
  } in !
```

NoamDev:
Previously, when Request interacted with a box, the box was instantiated
fixes #180


faiface (approved):
Thanks for catching this!

PR #184, opened 2026-04-03
Fix filename in packages_and_modules.md

AFelix5:
Corrected filename from 'posts.par' to 'post.par'


faiface (approved):
Good catch, thank you!

PR #186, opened 2026-04-10
Improve Handle performance

NoamDev:
This pull request transitions Handles to use mutex-based Variables instead of async oneshot channels.
And async oneshot channels are only used to read values.


faiface (approved):
LGTM, thank you so much!

PR #188, opened 2026-04-20
Some speedups

NoamDev:
I tried making some optimizations based on the following example:

```
module Main

import @core/Nat
import @core/Bool
import @core/List

type DiffList<t> = [List<t>]List<t>

dec QuickSort: [List<Nat>] List<Nat>

def QuickSort = [xs] 
  let diff: DiffList<Nat> = xs.unfounded.case {
    .end! => [x]x,
    .item(x) xs => {
        let (list1)list2 = Partition(xs, x) in
        let list1: DiffList<Nat> = list1.loop in
        let list2: DiffList<Nat> = list2.loop in
        [tail] list1(.item(x) list2(tail))
    }
  } in
  diff(.end!)


dec Partition: [List<Nat>, Nat] (List<Nat>) List<Nat>
def Partition = [list, x] 
  let list1: DiffList<Nat> = [xs]xs in
  let list2: DiffList<Nat> = [xs]xs in

  list.begin.case {
    .end! => (list1(.end!))list2(.end!),
    .item(y) list => {
      Nat.Compare(y,x).case {
        .less! => {
          let list1 = [xs: List<Nat>] list1(.item(y) xs) in
          list.loop
        },
        else _ => {
          let list2 = [xs: List<Nat>] list2(.item(y) xs) in
          list.loop
        }
      }
    }
  }

def ListNat_Equals= [xs : List<Nat>, ys: List<Nat>] type Bool in chan res {
    xs.begin
    if {
        xs is .item(x) xs and ys is .item(y) ys and Nat.Equals(x,y) => {
            xs.loop
        },
        xs is .end! and ys is .end! => {
          res <> .true!
        }
        else => {
          res <> .false!
        }
    }
}

def Main: ! = do {
  let list = Nat.Range(1,2000)
  list -> QuickSort

} in !
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2254,6 +2254,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f9fbbcab51052fe104eb5e5d351cf728d30a5be1fe14d9be8a3b097481fb97de"
 
 [[package]]
+name = "libmimalloc-sys"
+version = "0.1.46"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bc89deee4af0429081d2a518c0431ae068222a5a262a3bc6ff4d8535ec2e02fe"
+dependencies = [
+ "cc",
+]
+
+[[package]]
 name = "libredox"
 version = "0.1.3"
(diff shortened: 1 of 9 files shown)


faiface (approved):
LGTM

PR #196, opened 2026-05-07
Have less overhead in Json.{Encode,Decode}

coolreader18:
I noticed that there's some unnecessary conversion overhead between `JsonValue` and `serde_json::Value`, so I removed that by implementing `serde::{Serialize, Deserialize}` directly.

### DIFF
--- a/crates/par-builtin/src/builtin/json.rs
+++ b/crates/par-builtin/src/builtin/json.rs
@@ -1,12 +1,12 @@
 //package: core
-use std::{collections::BTreeMap, sync::Arc};
+use std::sync::Arc;
 
 use crate::builtin::list::readback_list;
 use arcstr::literal;
+use indexmap::IndexMap;
 use par_runtime::primitive::ParString;
 use par_runtime::readback::Handle;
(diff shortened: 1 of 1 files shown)


faiface:
Oh so there was a PR here!

@coolreader18 I don't know why, but this PR was not shown in PRs, but there was an increased PR count. This must've been it!

Looks like a good change, I'll review it shortly, but probably tomorrow.

NoamDev:
Loll, it really messed with me

faiface:
@coolreader18 Two questions/remarks here.

1. Why have two structures instead of one (`SerJsonValue` vs `DeJsonValue`)? Is there any particular advantage to that?
2. I'm not sure the switch from `BTreeMap` to `IndexMap` here is great. The maps generated from `Map.New` and `BoxMap.New` (and `FromList`) are documented to iterate in key comparison order. Of course, the map interface doesn't require that per se, but it might be a good invariant to uphold here too, for consistency.

NoamDev:
@faiface 
On the other hand, not using IndexMap, means you lose information from the JSON document. Namely the order of entries.

coolreader18:
Closing for now - I wanna chew on this more

PR #201, opened 2026-05-12
Change Nat to BigUint in helper methods dealing with nat

coolreader18:
This makes it easier to guarantee that there will never be a negative Nat.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3046,6 +3046,7 @@ dependencies = [
  "indexmap",
  "inventory",
  "num-bigint",
+ "num-integer",
  "num-traits",
  "par-core",
  "par-runtime",
--- a/crates/par-builtin/Cargo.toml
+++ b/crates/par-builtin/Cargo.toml
@@ -18,6 +18,7 @@ im = "15"
 indexmap = "2.7.0"
 num-bigint = "0.4.6"
 num-traits = "0.2.19"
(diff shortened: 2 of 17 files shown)


NoamDev:
Currently Nat is a subtype of Int, which is convenient. This wouldn't be possible if they were represented differently.

faiface:
Hey @coolreader18, thank you for the effort! But unfortunately, @NoamDev is right, the `Nat` being represented as `BigInt` instead of `BitUint` is so that the `Nat < Int` subtyping works without the need for any runtime conversion. Your encoding could be made to work as well, if at every single place where `Int` is expected, you'd also make it support `Nat` additionally. That would require a runtime conversion, which could be costly too. Making them the same representation allows for easy no-op subtyping.

coolreader18:
I don't think my PR actually changes that? To my understanding, that's in the `Number` type in par-runtime, right? That's still intact -- the main change here is that `Handle::nat()` returns a `BigUint` now, and `Handle::provide_nat` takes a `BigUint`, but the inner `flat` representation is unchanged.

faiface:
@coolreader18 Ohh, you are actually correct! @NoamDev we are wrong. This is totally backwards compatible, and an improvement to the API. Nice!

faiface (approved):
LGMT

PR #202, opened 2026-05-16
Fix broken code snippets in docs

EatChangmyeong:
The [Error Handling](https://par.run/book/quality_of_life/error_handling) chapter contained code that didn't compile because its `.poll` branch is now a keyword:

```
  × Syntax error
    ╭─[9:3]
  8 │   .close => Result<e, !>,
  9 │   .poll => Result<e, (a) self>,
    ·   ▲
 10 │ }
    ╰────
  help: invalid either/choice branches
        expected `}`
```

This is the only code I'm aware of that the new `poll` keyword breaks.

This PR updates the branch to avoid that keyword. I chose `.next` as its new name, which is consistent to `Sequence<a>` in the [Iterative](https://par.run/book/types/iterative) chapter.

### DIFF
--- a/docs/book.toml
+++ b/docs/book.toml
@@ -15,4 +15,4 @@ mathjax-support = true
 edit-url-template = "https://github.com/par-team/par-lang/edit/main/docs/{path}"
 
 [output.html.code.hidelines]
-par = "?"
+par = "# "
--- a/docs/src/nondeterminism/fan_pattern.md
+++ b/docs/src/nondeterminism/fan_pattern.md
@@ -8,7 +8,7 @@ module Main
 import @core/List
 
 dec MergeTwoLists : <a>[List<a>] [List<a>] List<a>
-def MergeTwoLists = [left, right] poll(left, right) {
+def MergeTwoLists = <a>[left] [right] poll(left, right) {
   list => list.case {
     .end! => submit(),
     .item(x) xs => .item(x) submit(xs),
--- a/docs/src/nondeterminism/poll_submit.md
+++ b/docs/src/nondeterminism/poll_submit.md
@@ -82,7 +82,7 @@ dec PollSum : [List<Int>] Int
 def PollSum = [nums] poll(nums) {
   list => list.case {
     .end! => submit(),
-    .item(x) xs => x + submit(xs),
+    .item(x) xs => x + {submit(xs)},
   }
   else => 0,
 }
@@ -126,7 +126,7 @@ dec PollSumTwo : [List<Int>, List<Int>] Int
 def PollSumTwo = [nums1, nums2] poll(nums1, nums2) {
   list => list.case {
     .end! => submit(),
-    .item(x) xs => x + submit(xs),
+    .item(x) xs => x + {submit(xs)},
   }
   else => 0,
 }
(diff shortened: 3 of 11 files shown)


EatChangmyeong:
I've identified more broken code snippets which I decided to merge into this PR:

* [Primitive Types](https://par.run/book/structure/primitive_types)
  * Add new definitions `Yes` and `No` that are explicitly typed `Bool`, since logical operators doesn't work with `either { .true! }` nor `either { .false! }`
  * Add missing `import @core/Nat` that breaks `@core/Int` example
* [The `let` Expression](https://par.run/book/structure/let_expressions): add missing trailing unit
* [Either](https://par.run/book/types/either): remove extraneous comma
* [Forall](https://par.run/book/types/forall): add missing `import @core/Option` that breaks the second snippet. I'm aware that it's supposed to refer to `type Option` earlier, but using the built-in definition makes more sense for me. I've also added an explanation that `Option` is built-in. (Please rewrite that part if it sounds off/wrong/awkward!)
* [Recursive](https://par.run/book/types/recursive): replace `Int` with `a`
* [Channels & Linking](https://par.run/book/processes/chan_expression): Add new local variable `isZero`, since `{n == 0}.case` triggers syntax error
* [Construction by Destruction](https://par.run/book/processes/duality): fix `dual ! = ?` rendering, where `?` on the right-hand side was interpreted as [line-hiding prefix](https://rust-lang.github.io/mdBook/format/mdbook.html#hiding-code-lines)
  * I've changed `output.html.code.hidelines.par` in `book.toml` from `"?"` to `"# "`, which is consistent to Rustdoc's convention. As far as I know, the only place Par uses the `#` character is [before generated variable names in the compiled view](https://par.run/book/process_syntax#what-even-is-a-process-language).
* [Error Handling](https://par.run/book/quality_of_life/error_handling): replace every occurrence of `path.createOrAppendToFile` or similar with `Os.CreateOrAppendToFile(path)`
* [Pipes](https://par.run/book/quality_of_life/pipes): replace piped `3` with `{3}`
* [Polling & Submitting](https://par.run/book/nondeterminism/poll_submit)
  * Replace `submit(xs)` used as operand with `{submit(xs)}`
  * Replace non-generic `[left, right]` binder in `MergeTwoLists` with `<a>[left] [right]`
* [The Fan Pattern](https://par.run/book/nondeterminism/fan_pattern): replace non-generic `[left, right]` binder in `MergeTwoLists` with `<a>[left] [right]`

Speaking of line-hiding prefix, I also think that `module Main import { modules... }` lines should be hidden, and more code blocks should have hidden boilerplate lines so that just copy-pasting the code into the IDE just works, but that doesn't really belong to this PR.

faiface (approved):
Very good job! Thank you so much @EatChangmyeong.

I am approving, but some of these I'd actually consider parser bugs / deficiencies.

I'll make another issue for them, gonna link it afterwards.

faiface:
Here's the new issue: https://github.com/par-team/par-lang/issues/204

faiface:
@EatChangmyeong I'm good with merging this as is and tackling the other issue separately. What do you think?

EatChangmyeong:
I'm fine with that. One last thing to mention: should the docs mention that [`AddBothOrZero` is currently not working and we're aware of it](https://github.com/par-team/par-lang/issues/203)?

faiface:
@EatChangmyeong I don't think that's worth it mentioning in the docs, considering we'll most likely fix it soon.

Great, I'm gonna merge this then!

PR #205, opened 2026-05-18
If there's an error, only report that error

coolreader18:
Previously, if you typoed one name, every single identifier in the workspace would be marked as an error. Now, instead of throwing away the `TypeDefs` if there's an error in `TypeDefs::new_with_validation`, we keep it and just report the errors in addition.


faiface (approved):
Oh thank you so much! This has been bugging me for a while, but I've never gotten around to fixing it.

Great stuff! Gonna merge right away.

PR #206, opened 2026-05-20
Improve dev experience for builtin packages

coolreader18:
`par check --package ./crates/par-builtin/packages/{core,basic}` now properly recognizes that these are dev versions of the builtin packages, and can successfully compile them. In the second commit, I remove the duplication of `PackageId` and `PackageID`. If it's desired that I split this out into a separate PR, I could do that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2260,7 +2260,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "fc2f4eb4bc735547cfed7c0a4922cbd04a4655978c09b54f1f7b228750664c34"
 dependencies = [
  "cfg-if",
- "windows-targets 0.48.5",
+ "windows-targets 0.52.6",
 ]
 
 [[package]]
@@ -3087,6 +3087,7 @@ dependencies = [
  "miette",
  "par-builtin",
  "par-core",
+ "par-runtime",
  "pathdiff",
  "percent-encoding",
  "pulldown-cmark",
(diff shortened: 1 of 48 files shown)


faiface (changes requested):
Hi @coolreader18! This is an extremely useful and highly welcome change! Thank you so much.

I've just one problem: I think `PackageId` belongs in `par-core`, not `par-runtime`. With that, you can also get rid of the `par-runtime` dependency in `par-doc`.

After that, I think this is good to merge.

coolreader18:
The issue is that `par_runtime::linker` uses `PackageId`, and already had the `PackageID` type within it.

faiface:
@coolreader18 I see, but the other one that you removed was in core, no? So, I'd still argue that it should be in core. Regardless of how it was, that's where it belongs.

faiface:
Oh, I see your argument actually. With that you'd need to add a core dependency to runtime, which is not desired.

faiface:
That was actually the reason for the duplication of PackageId. To avoid that dependency, but still have PackageId in core so that places like doc don't have to depend on runtime. Now that I think about it, it probably makes sense to restore that duplication.

coolreader18:
The core crate already depends on runtime though, so if doc depends on core it already transitively depends on runtime.

coolreader18:
This could also just be solved by making a new `common` crate.

faiface:
@coolreader18 Hm, good point. I'll need to have a closer look to figure out if a common crate or a different solution is suitable.

So, for now, I think your solution is okay. I'm good with merging this.

Okay with you?

faiface (approved):

coolreader18:
Sounds good!

PR #207, opened 2026-05-20
Use bytes_utils::Str for ParString

coolreader18:
This eliminates the redundant utf-8 checks for ParString, since `bytes_utils::Str` handles the validity invariant for us.


faiface (approved):
This looks good to me!

Thank you @coolreader18

PR #211, opened 2026-06-03
Use `string_cache` instead of `arcstr` for signals and identifiers

coolreader18:
This allows constant-time comparison for things like `match handle.case().await {}`, as well as preventing the likelihood of misspellings.

Perhaps `string_cache` isn't the perfect library for this, and perhaps we'd want to move to a bespoke implementation in the future, but for now I think it's better than the current solution. I was mainly looking for something akin to rustc's `rustc_span::symbol` but that implementation isn't reusable without copy-pasting.


faiface:
I'll need to take a closer look at this one.

NoamDev:
Nice! This has been bothering me for quite a while.
It also reduces the size of signals in the runtime which used to be `(usize,usize)`, and would now be `Atom` which is u64.

The only thing that bothers me is `crates/par-runtime/src/atom/sym.rs` since it would be annoying to maintain.

I think it'd be nicer if those symbols were declared in the relevant `builtin` files, which would make it more localized.

NoamDev:
I'm also unsure about how Atom serializes.
It basically creates many copies of the same string, and then unifies them at deserialization which is pretty wasteful.
Also, the Atoms which aren't static are ref-counted, which would cause overhead when they're copied/destroyed.

So like the interface is great, but we might want to implement something like that ourselves instead 😅

faiface:
@coolreader18 I'm going to close this, Noam has partly reimplemented some of these changes, but overall, I don't think the `sym` idea is a good fit at this point, as it requires hardcoding the symbols and code generation, complicating the build process, for not that much performance or correctness benefit. I say not much correctness benefit because it is not hard to validate the `literal!(...)` either by looking at them (there's no logic, they's either a typo or there's not), or simple unit tests covering the path where the symbol gets invoked: it either always crashes, or it doesn't. We've measure the performance benefit and it's quite negligible.

However, thank you so much for the effort here, it's very much appreciated, and I'll be happy to see more contributions from you! Thank you so much :)

PR #212, opened 2026-06-03
Miscellaneous docs improvements

coolreader18:
* Render choices and eithers more like they'd typically be written (`.none!`, `.add(String) => self` rather than `.none !`, `.add => [String] self`)
* Order primary exports (String.String, Map.Map) first in their module pages
* Add docs for primitive types


faiface (changes requested):
Oh beautiful! You really have a good instinct for finding things that I was planning to do later and now I don't have to :D 

One detail: I think the more standard phrase is "arbitrary-precision" instead of "unlimited-precision". With that, I'm ready to merge this.

coolreader18:
Ah, yep, that was the wording I was trying to think of :)

faiface (approved):
Great, thank you so much! Merging this :)

PR #221, opened 2026-06-25
 typechecking: infer .begin.case result type in chan blocks

### LINKED ISSUE #218: `.begin` fails type inference in a `chan` block
I'm trying to learn Par through the book. When reading [this](https://par.run/book/quality_of_life/error_handling#propagating-errors-in-functions), I wanted to check that `return <> .ok contents` is the same as `return.ok <> contents`; however, when I tried to run this locally, it did not compile:
```par
module Main

import {
  @basic/Os
  @core/Bytes
  @core/Try
}

dec ReadAll : [Os.Path] Try<Os.Error, Bytes>
def ReadAll = [path] chan return {
  catch e => { return <> .err e }
  let try reader = Os.OpenFile(path)
  let parser = Bytes.ParseReader(reader)
  let try contents = parser.begin.case {
    .empty! => .ok <<>>,
    .some parser => parser.remainder,
  }
  return <> .ok contents
}
```

```
# ~/Projects/hello_par 
; par check                   
Checking package: .
  × Type of `#result` must be known at this point.
    ╭─[14:28]
 13 │   let parser = Bytes.ParseReader(reader)
 14 │   let try contents = parser.begin.case {
    ·                            ──────
 15 │     .empty! => .ok <<>>,
    ╰────

  × Type of `#result` must be known at this point.
    ╭─[14:34]
 13 │   let parser = Bytes.ParseReader(reader)
 14 │   let try contents = parser.begin.case {
    ·                                  ─────
 15 │     .empty! => .ok <<>>,
    ╰────
```

The fact that the identifier it complains about is something generated during desugaring doesn't make it easier to understand. I figured out I need to replace `parser.begin` with `{type Bytes.Parser<Os.Error> in parser.begin}`, though… which sounds like something the compiler should already understand.

What I think is an MRE of the same issue:
```par
module Main

import {
  @core/Option
  @core/List
}

dec Head : <a: box> [List<a>] Option<a>
def Head = <a: box> [items] chan return {
    let res: Option<a> = items.begin.case {
        .end! => .none!,
        .item(x) _ => .some x
    }
    return <> res
}
```
I checked this latter snippet on the playground to make sure the issue isn't caused by me using too old of a version (I use the version from Nixpkgs, which builds from the commit 56ecfe8).

notJoon:
## Description

Closes #218 

Fixed `begin.case` so it can infer the result type without an annotation when the recursive subject's type is already known

### DIFF
--- a/crates/par-core/src/frontend_impl/types/checking.rs
+++ b/crates/par-core/src/frontend_impl/types/checking.rs
@@ -1611,7 +1611,12 @@ impl<S: Clone + Eq + std::hash::Hash> Context<S> {
         mode: &ProcessAnalyzerMode,
         emit: &mut impl FnMut(TypeError<S>),
     ) -> (Command<Type<S>, S>, Option<Type<S>>) {
-        if let Some(inference_subject) = inference_subject {
+        // when inferring another subject, .begin is only problematic if this
+        // object's type is still unknown. known recursive types
+        // can be checked normally.
+        if let Some(inference_subject) = inference_subject
+            && !matches!(typ, Type::Recursive { .. })
+        {
             emit(TypeError::TypeMustBeKnownAtThisPoint(
                 span.clone(),
                 inference_subject.clone(),
@@ -1693,7 +1698,7 @@ impl<S: Clone + Eq + std::hash::Hash> Context<S> {
         if let Err(e) = self.put(span, object.clone(), expanded) {
             emit(e);
         }
-        let (process, _inferred_type) = self.analyze_process(process, mode, emit);
+        let (process, inferred_type) = self.analyze_process(process, mode, emit);
         (
             Command::Begin {
                 unfounded,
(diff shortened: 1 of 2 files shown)


NoamDev:
Hey @notJoon ,

The reason we don't infer the result type after a `begin`, is that in general it is a hard problem.

Consider this simple example:

```
def A : List<!> =
let list: List<!> = *(!,!,!) in
list.begin.case{
  .end! => .end!,
  .item(x) list => .item(x) list.loop
}
```

Here we'd need to infer the result type as `List<!>`, which would require solving recursive equations.

In particular, in your PR this creates some unclear error.

Also, note that none of your tests include `loop`. If you're not using loop, the `begin` keyword is redundant. And this way it'll all work without any changes.

For example:

```
dec Head : <a: box> [List<a>] Option<a>
def Head = <a: box> [items] chan return {
  let res = items.case {
    .end! => .none!,
    .item(x) _ => .some x,
  }
  return <> res
}
```

would work perfectly on master.

NoamDev:
Closing for now.

NoamDev:
Feel free to respond/contest tho!

PR #222, opened 2026-06-25
docs: use `Bytes.ReadAll` in error handling snippet

notJoon:
While working on #221, I found a docs snippet that failed `par check` with:

```text
Type of `#result` must be known at this point.
```

This updates the snippet to call `Bytes.ReadAll(reader)`, which already returns `Try<Os.Error, Bytes>` and avoids the problematic `#result` type inference point.

### DIFF
--- a/docs/src/quality_of_life/error_handling.md
+++ b/docs/src/quality_of_life/error_handling.md
@@ -355,8 +355,8 @@ Since `try` is part of the pattern, it works in nested patterns too:
 
 ```par
 let (try leftReader, try rightReader)! = (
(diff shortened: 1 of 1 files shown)


NoamDev:
As said elsewhere, the solution here is to omit the `.begin`.

faiface:
@NoamDev That's true, but this change overall is an improvement to the example considering the current standard library functionalities. So I'm good with it and gonna merge this.

Thanks @notJoon !

faiface (approved):

PR #223, opened 2026-06-26
Add autocomplete

Toucan4Life:
Hi everyone, continuing with my DX Improvements.

This was one of my biggest pain point when programming in vscode, not knowing what was the method behind a type or what you could do in a loop / recursive context.

The PR seems huge, but there is more than 1000 lines of test, I tried to be as exhaustive as possible, but I am pretty sure some suggestions slipped through.

I mark this as draft, because even though I think it is complete, I completely understand that this PR is huuuge, and I am totally open to splitting into multiple PRs to make the review easier. I also tried to reuse some typecheck / parser logic,, but couldn't wrap my head around it

/AI Disclaimer\: AI was using to write the tests (to preserve my sanity with all the edge cases) and some of the comments (since english is not my first language).

What different king of completion it is capable of:

- On a Choice: 
image

-  On an Either constructor:
image

-  On an Either:
image

- On Recursive:
image

- On Module: 

image

I'll focus more on the label and detail of the CompletionCandidate (basically have better description and example in the completion) on another PR, otherwise the scope would have been to big

### DIFF
--- a/crates/par-core/src/workspace.rs
+++ b/crates/par-core/src/workspace.rs
@@ -29,7 +29,7 @@ use std::sync::Arc;
 
 mod completion;
 
-pub use completion::CompletionCandidate;
+pub use completion::{CompletionCandidate, CompletionCandidateKind};
 
 pub type ExternalModule = Module<Arc<process::Expression<(), Unresolved>>, Unresolved>;
 
--- a/crates/par-core/src/workspace.rs
+++ b/crates/par-core/src/workspace.rs
@@ -27,6 +27,10 @@ use std::fs;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
 
+mod completion;
+
+pub use completion::CompletionCandidate;
+
 pub type ExternalModule = Module<Arc<process::Expression<(), Unresolved>>, Unresolved>;
 
 const MANIFEST_FILE: &str = "Par.toml";
--- a/crates/par-core/src/workspace/completion.rs
+++ b/crates/par-core/src/workspace/completion.rs
@@ -2,7 +2,7 @@ use super::CheckedWorkspace;
 use crate::frontend_impl::language::{LocalName, Universal};
 use crate::frontend_impl::types::Type;
 use crate::location::FileName;
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashSet};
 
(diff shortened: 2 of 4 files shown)


faiface:
Whoa! Thanks so much for this! I'll review this asap, tho expect a couple of days.

This is wonderful!

faiface:
One question. Does it also work right after typing a `.` in a context? I.e. in a state that's not valid syntax?

If not, there should be a fairly easy way to solve that: Add a special Cursor token at the place of the cursor (if the cursor exists), and parse it as one of the expected things (not gonna type check, but gonna parse), from which completion can flow.

Toucan4Life:
Depends where. 

Not in this case, where we have subbranch inside an either or a choice : 
image
but as a drill down of a known class or module yes ! :) 
image

faiface:
Tried it out, it works really well! I'm impressed.

Apologies for slower responses, we're working on a paper and currently nearing a deadline, so I hope you understand.

Toucan4Life:
Don't worry at all! It is usually hearing no feedback from maintainer that demotivate would be contributor and here this is not the case :)

Take all the time you need, real life stuff are obviously more important. And again, if you want me to split up the PR to be easier to review, don't hesitate to  tell me so. In the future I will try to keep them smaller.

Good luck on your paper !

faiface:
@Toucan4Life I'm back! I like this change. It's of course not the most powerful auto-complete imaginable, but it's good enough for the time being! I'm good turning it to a proper PR (non-draft) and merging it. Is there something you'd like to change there before we do?

Toucan4Life:
Nice to hear! Yeah I will work on completing the autocomplete (see what I did there) in future PR, I am not done yet.

For this PR I am ok merging it as is, like you said I find it a good start

faiface (approved):

faiface:
Great, merging!

Toucan4Life:
Thanks!

My goal is to have a complete (ah!) autocomplete, , so I also want to add  :
 -better description
 -example when navigating the autocomplete suggestions
 -more cases where the autocomplete trigger ( like the subbranch example above)

On what do you want me to start first? Or on something else?

Toucan4Life:
@faiface ping on the message just above :) BTW I hope your paper went well !

PR #224, opened 2026-07-13
refactor: update Arena struct to use Vec<ArcStr> for strings

NoamDev:
Now String indexes are a single `usize`.


faiface (approved):
LGTM :)

PR #225, opened 2026-07-14
refactor(builtin): replace all `…_…_external!` macros with a single macro

magistau:
The macro is available at `par_runtime::external_def!`. Usage syntax:
```rust
external_def! {
    @package/path/to/Module.Name => handler
}
external_def! {
    @package/path/to/Module.Name => handler(arg1, arg2)
}
external_def! {
    @package/path/to/Module.{
        Name1 => handler1,
        Name2 => handler2(arg1, arg2),
    }
}
```

### DIFF
--- a/crates/par-builtin/src/builtin/bench.rs
+++ b/crates/par-builtin/src/builtin/bench.rs
@@ -1,21 +1,9 @@
+use par_runtime::external_def;
 use par_runtime::readback::Handle;
-use par_runtime::registry::{DefinitionRef, ExternalDef, PackageRef};
 
-macro_rules! core_bench_external {
-    ($name:literal, $f:path $(, $arg:expr)*) => {
-        inventory::submit!(ExternalDef {
-            path: DefinitionRef {
-                package: PackageRef::CORE,
-                path: &[],
-                module: "Bench",
-                name: $name,
-            },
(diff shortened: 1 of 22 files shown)


NoamDev:
Very pretty, love it!
One issue tho, this should also be supported
```
external_def! {
    @package/path/to/Module.Name => handler
}
```

And is exactly what `path: &[]` is meant for.

magistau:
@NoamDev would your example have this?
```rust
DefinitionRef {
    package: todo!("package ref for `@package`"),
    path: &["path", "to"],
    module: "Module",
    name: "Name",
}
```
I'm not very experienced with neither Par nor its implementation, and I haven't seen any examples of `path` being different from `&[]`.

NoamDev:
Yes, exactly!
And yeah the standard library is quite flat, but we support more hierarchy.

If you're curious (tho I don't think it's necessary for this PR), you can read about the module system in [the docs](https://par.run/book/structure/packages_and_modules).

You can also take a look at the [original design](https://github.com/par-team/par-lang/issues/133), tho it doesn't exactly match the actual implementation.

magistau:
@NoamDev I've updated the macro as you suggested and also added some documentation.

faiface (approved):
Beautiful stuff! LGTM

PR #230, opened 2026-07-18
Change implicit generic syntax from `<a>[T] ...` to `[<a> T] ...`

faiface:
Currently, the syntax for implicit generics is like this:
```
dec Map : <a>[List<a>] <b>[box [a] b] List<b>
def Map = <a>[list] <b>[f] ...
```
The generic parameter (such as `a`) is inferred only from the following argument (such as `List<a>`). But written outside of the parentheses like this forces the two arguments to be separated into separate pairs of parentheses.

If we moved the `<a>` inside the parentheses instead, it would still be clear where the generic parameter is inferred from, while using a single pair of parentheses:
```
dec Map : [<a> List<a>, <b> box [a] b] List<b>
def Map = [<a> list, <b> f] ...
```
Notice especially, that the `<a>` and `<b>` still appear in the `def`, but before the argument names.

If we then wrote type annotations inside the `def`, it would look like this:
```
def Map = [<a> list: List<a>, <b> f: box [a] b] ...
```

### DIFF
--- a/crates/par-builtin/packages/basic/src/Sql.par
+++ b/crates/par-builtin/packages/basic/src/Sql.par
@@ -118,11 +118,11 @@ export {
   // Decode an optional value with a format. Missing columns return `.err`.
   // Designed for pipe chains:
   //   row->Sql.Field("age")->Sql.Decode(Sql.AsInt)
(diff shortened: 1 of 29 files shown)


NoamDev:
There are a few places you missed:
image

PR #231, opened 2026-07-19
refactor: parser

magistau:
- fix `list0` allowing `,` to be parsed (that is, `*(,)` is parsed as an empty list)
- use existing interfaces like `delimited` from `winnow` or `Option::unwrap_or_else` from `core` to reduce boilerplate
- fix certain issues `clippy` sees in this module
- ignore the results of certain parsers using `preceded`, `terminated`, `delimited`, `separated_pair` and `seq!`, thus removing the need to match them as unused bindings
  - introduce `commit_prec` and `commit_delim`, versions of `commit_after` allowing ignoring the prefix and optionally a suffix
- use some syntax sugar (elided lifetimes/types, matching binding names with struct field names to replace `foo: bar` with `foo` etc.)
- move out the bodies of parsers for `begin` and `unfounded` to parametrized functions to reduce boilerplate

### DIFF
--- a/crates/par-core/src/frontend_impl/parse.rs
+++ b/crates/par-core/src/frontend_impl/parse.rs
@@ -119,10 +119,7 @@ where
     P: Parser<Input<'i>, O, Error>,
     Vec<O>: Accumulate<O>,
 {
-    terminated(
-        separated(0.., item, t(TokenKind::Comma)),
-        opt(t(TokenKind::Comma)),
-    )
+    opt(list1(item)).map(Option::unwrap_or_default)
 }
 
 fn list1<'i, P, O>(item: P) -> impl Parser<Input<'i>, Vec<O>, Error> + use<'i, P, O>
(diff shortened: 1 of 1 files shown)


NoamDev:
Nice, it'd be good if you can add a test for that. That `*(,)` for example does not parse.

faiface:
@magistau Hey, I'm good merging this, are you going to add smth else? If not, let's just merge it.

magistau:
@faiface I've pushed everything I had for now. I can always send a new PR if I have something else, so this is ready for review.

faiface:
Great! I like lines going down :D @magistau Can you briefly summarize the intents behind the changes so it's easier to review?

magistau:
@faiface I've updated the PR description.

faiface:
@magistau Great, this looks good to me! Thank you so much.

faiface (approved):

PR #234, opened 2026-07-26
Implement typed holes (todo)

### LINKED ISSUE #187: Feature request: typed holes
Often you might want to type check a partially constructed program. In Rust you can insert `todo!()` in place of an expression and in proof assistants you can often insert a *hole* in place of an expression. It would be great to have typed holes in Par too.

NoamDev:
This PR adds a process terminator `todo`, and a matching expression.
Those can be used to leave code unfinished while letting the type checker do its work.

During type checking we make sure there are no variables being inferred since there's no way to know what they're supposed to be.

Further, we always show a warning that the definition is incomplete.

Lastly, definitions which contain `todo` directly or indirectly do not appear in the final binary. If one attempts to run them, they'd get an error directing them to run `par check`.

Example usage:

`def T:! = todo`

`def T:! = chan res {todo}`

### DIFF
--- a/crates/par-core/src/api.rs
+++ b/crates/par-core/src/api.rs
@@ -101,6 +101,7 @@ pub mod frontend {
 }
 
 pub mod runtime {
+    pub use crate::backend::flat::transpiler::{Transpiled, TranspiledGlobal};
     pub use crate::runtime_impl::{Compiled, RuntimeCompilerError};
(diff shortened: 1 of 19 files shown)


faiface (approved):
Beautiful! I love how the compilation recursively folds definitions containing `todo` into "unimplemented", so we only get fully implemented stuff compiled, and thereby we are allowed to run definitions that don't reference `todo`, but not those that do! Great code all around, merging!

PR #237, opened 2026-07-31
Support dot completion inside case branches

Toucan4Life:
Add dot-autocomplete for `case` branch heads in Par completion.

## `either` branch heads

Before, typing inside a branch head like this did not propose the valid branches:

```par
type Stream = either {
  .end!,
  .chunk !,
}

def StreamValue : Stream = external
def Main : ! = StreamValue.case {
  .
}
```

Now it suggests:

- `chunk()`
- `end!`

## `choice` branch heads

For `begin case`, branch-head completion now works here:

```par
type Sequence = iterative choice {
  .close => !,
  .next => self,
}

def Main : Sequence = begin case {
  .
}
```

and suggests:

- `close`
- `next`

## Recursive / named types

This also works when the branch set comes from a recursive or named type rather than from text already present in the current `case` block.

Example:
```par
type Repeat = recursive either {
  .end!,
  .step self,
}

def RepeatValue : Repeat = external
def Main : ! = do {
RepeatValue.begin.case {
  .
}
} in !
```

Now it suggests:

- `end!`
- `step()`

Anything that might be missing?

My next target is to work on the description of the autocomplete, and when the todo keyword is implemented, I would like that when choosing a suggestion from the autocomplete automatically insert some code snippet. Like for the branch that this PR worked on, instead of inserting only .step(), it would insert .step() => todo.

I would also like to work on the hover tooltip; I would like that when you hover a function or a reserved keyword of par, it show an hover of the usage of that function/keyword, like in rust.

Or do you want me to work on another things for my next PR?

Also congrats on the paper :D

### DIFF
--- a/crates/par-core/src/workspace/completion.rs
+++ b/crates/par-core/src/workspace/completion.rs
@@ -31,28 +31,49 @@ impl CheckedWorkspace {
         let Some((dot, hover_row, hover_column)) = dot_before_position(source, row, column) else {
             return Vec::new();
         };
-        let Some(completion_context) = DotCompletionContext::before_dot(source, dot) else {
-            return Vec::new();
-        };
-        let active_loop_labels = loop_labels_before_dot(source, dot);
 
         let mut candidates = Vec::new();
-        let completed_module_alias =
-            self.push_module_alias_completion_candidates(file, source, dot, &mut candidates);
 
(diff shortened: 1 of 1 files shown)


faiface:
Nice, this looks good, thank you @Toucan4Life, merging!

faiface (approved):

